Issue #15, opened 2018-11-13
Update SVG icons 


milaendo:
replace old icons with new font awesome SVGS

PR #17, opened 2018-11-14
new icons 

milaendo:
screen shot 2018-11-14 at 10 32 46 am
#15


velkymx:
@milaendo can we change the first one to say "Text box" instead of Text?

PR #19, opened 2018-11-16
Ability to add color to text field

milaendo:
In the inspector there is now a section for color which will accept all html colors and hex codes
#18


milaendo:
screen shot 2018-11-16 at 10 09 58 am

PR #29, opened 2018-12-06
add name and value to submit button

marcoAntonioNina:
- Add change value in button submit
- value in event click


velkymx:
@marcoAntonioNina the submit and prompt happens BEFORE the data changes. The object should be updated before that.

PR #31, opened 2018-12-11
Add support for computed properties as a form property

marcoAntonioNina:
Feature https://github.com/ProcessMaker/vue-form-builder/issues/22

- Continuing with Taylor's work, only the new variable is established.
- the library used is https://www.npmjs.com/package/expr-eval
properties
image

result
image


velkymx:
@marcoAntonioNina editing doesn't work. It says you can't have a duplicate. Also the button should say update or something instead of add property on edit.

Issue #34, opened 2018-12-18
If no fields are filled, the form's submit sends null: null


caleeli:
Related to: https://github.com/ProcessMaker/bpm/issues/986

If the check was not checked it did not 
![image](https://user-images.githubusercontent.com/8028650/50158448-8a547f00-02aa-11e9-99a0-1ae990e96501.png)

![image](https://user-images.githubusercontent.com/8028650/50158691-31d1b180-02ab-11e9-994b-5ec06eaf7483.png)

PR #36, opened 2018-12-18
Send default values when submit a form. #34

caleeli:
Related to #34 
The following default values are sent if the controls are not activated or changed:

- Line input = empty string
- Select = null
- Radio group = null
- Checkbox = false
- Textarea = empty string
- Date picker = null
- Record List = empty array

image

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -78,6 +78,11 @@ export default {
     FormTextArea
   },
   watch: {
+    mode(mode) {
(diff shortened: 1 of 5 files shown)


caleeli:
@velkymx fixed the issue with the multicolumns when delete an item fuction was not implemented yet.
And refresh the data when click of preview, so the new names will be updated in the output data.

PR #42, opened 2019-01-30
Enabling datetimepicker in multicolumns # 41

danloa:
The datetimepicker wasn't registered for multicolumns, the component was imported and registered and now is working properly.


ryancooley:
@danloa I'm having trouble when running this integrated into BPM. Any idea what would cause this error?

screen shot 2019-01-30 at 11 00 39 am

PR #53, opened 2019-02-08
Use the field name as a css class selector

nolanpro:
Allows you to use the field name as a selector for custom css, for example an input with the field name `foo` can be used like `.foo { color: red }` in the custom css.

Also added a specific css class name field for text box since it doesn't have a field name

fixes: https://github.com/ProcessMaker/vue-form-builder/issues/55


velkymx:
This should add the ID to the div of the element. Then use that to access the element with the #id syntax

nolanpro:
Closing to reopen - wrong base branch

PR #57, opened 2019-02-11
Set the wraper ID to the field name so CSS can use it

### LINKED ISSUE #55: Add ID to Element
As a developer I want to be able to access individual form elements using the ID of the surrounding div. 

Change the code to use the Field Name of the element as the ID of the element div.

nolanpro:
Fixes #55 

Allows you to use the field name as the wrapper ID for custom css, for example an input with the field name `foo` can be used like `#foo { font-size: 40px }` in the custom css.

Also added a specific css class name field for text box since it doesn't have a field name

Also added Monaco editor as part of this PR

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
   ],
   "dependencies": {
     "@fortawesome/fontawesome-free": "^5.6.1",
-    "@processmaker/vue-form-elements": "git://github.com/ProcessMaker/vue-form-elements.git#bugfix/10",
+    "@processmaker/vue-form-elements": "^0.3.0",
     "bootstrap": "^4.1.3",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -18,23 +18,24 @@
   ],
   "dependencies": {
     "@fortawesome/fontawesome-free": "^5.6.1",
-    "@processmaker/vue-form-elements": "^0.2.3",
+    "@processmaker/vue-form-elements": "git://github.com/ProcessMaker/vue-form-elements.git#bugfix/10",
```
velkymx: @nolanpro this will update everyone's branch. Please remove.

PR #58, opened 2019-02-12
Inspector Scrolling #56

### LINKED ISSUE #56: Inspector Scrolling
The inspector needs to be able to scroll. Current content is off the screen.

marcoAntonioNina:
Fixes #56 

- Add style for scroll in inspector.

### DIFF
--- a/src/components/vue-form-builder.vue
+++ b/src/components/vue-form-builder.vue
@@ -56,11 +56,11 @@
       </div>
     </div>
 
-    <div class="inspector-container">
+    <div class="d-flex flex-row flex-column">
       <div class="card-header">
         Inspector
       </div>
-      <div class="card-body editor-draggable">
+      <div class="card-body">
         <component v-for="(item, index) in inspection.inspector" :formConfig="config" :key="index" :is="item.type" v-bind="item.config" v-model="inspection.config[item.field]" />
       </div>
     </div>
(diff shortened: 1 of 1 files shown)


src/components/vue-form-builder.vue
```diff
@@ -269,7 +269,14 @@ export default {
     width: 340px;
     max-width: 340px;
     border-left: 1px solid #e9edf1;
+    overflow: hidden;
+    padding-bottom: 5%;
```
tdondich: NO NO NO.  Utilize flexbox.  Do NOT use percentages like this.

src/components/vue-form-builder.vue
```diff
@@ -269,7 +269,14 @@ export default {
...
+  }
+
+  .inspector-container > .card-body {
     overflow: auto;
+    height: 95%;
+    padding-bottom: 20px;
```
tdondich: HORRIBLE.  Do not make assumptions on the overall layout dimensions.  Utilize flexbox.

src/components/vue-form-builder.vue
velkymx: The idea behind the bottom padding is to give a little extra spacing for the scroll.

tdondich: Percentages, ideally, should be avoided as it makes unsafe assumptions regarding viewport dimensions.  You can use calc() css statements if needed to do math more accurately.  Ideally, padding should also be based on 4s.  So 4px, 8px, 16px, 32px.

Issue #73, opened 2019-02-22
Preview Not Updating


velkymx:
When I add a record list and a form control, the form control is blank in the preview. I've also noticed that the preview stops updating at some point.

PR #81, opened 2019-02-25
Fix warnings in form builder - Record List

caleeli:
image

Related: #73 , #76


caleeli:
Looks like navigation buttons are also working.
image

PR #91, opened 2019-02-26
Use the wysiwyg control in the builder

### LINKED ISSUE #74: Move WYSIWYG
The WYSIWYG control should ONLY work on the text field. Move the WYSIWYG to a form-elements object and import it as it's own type. Revert the change to the text object. 

https://github.com/ProcessMaker/vue-form-elements/issues/20

nolanpro:
This depends on https://github.com/ProcessMaker/vue-form-elements/pull/24. Once released we will need to updated the version of vue-form-elements in this PR

Fixes #74 

- Update how we handle draggable active highlight elements work since we don't always have a "mask" to do that job. Interactive elements can not have masks.
- Update click events so they show the correct inspector


src/components/editor/html-editor.vue
```diff
@@ -0,0 +1,50 @@
...
+  <div class="highlight form-group" @click.stop="$emit('inspect')">
+    <div class="draggable-handle">
+      <i class="fas fa-arrows-alt"></i>
+    </div>
+    <form-html-editor name="htmlEditor" @onFocus="$emit('inspect')" label="Html Editor" v-model="content" />
+  </div>
+</template>
+
+<script>
+import { FormHtmlEditor } from "@processmaker/vue-form-elements/src/components";
```
chipit24: I know this branch is just for testing, but elements should be imported from `@processmaker/vue-form-elements`. Reaching directly into `src/components` means we bypass the bundled components built with `npm run build-bundle`.

PR #95, opened 2019-02-28
Scrolling Screwed Up

### LINKED ISSUE #79: Scrolling Screwed Up
Each of the three columns should flex and scroll AS NEEDED and INDEPENDENTLY. The toolbar should be locked to the top as well as the footer.

marcoAntonioNina:
Fixes #79 

- Add scroll panels

### DIFF
--- a/src/components/vue-form-builder.vue
+++ b/src/components/vue-form-builder.vue
@@ -269,7 +269,7 @@
                 padding: 4px;
 
                 .control {
-                    margin: 0 8px;
+                    margin: 8px;
                     width: 100%;
                     height: 32px;
                     display: flex;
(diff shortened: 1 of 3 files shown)


src/components/vue-form-builder.vue
```diff
@@ -1,389 +1,393 @@
...
-        :options="{sort: false, group: {name: 'controls', pull: 'clone', put: false}}"
-        :clone="cloneControl"
-      >
-        <div class="control" v-for="(element, index) in controls" :key="index">
-          <div class="icon">
-            <img v-if="element['editor-icon']" :src="element['editor-icon']">
-            <i v-if="element['fa-icon']" :class="element['fa-icon']"></i>
-          </div>
-          <div class="label">{{element.label}}</div>
+    <div class="dynaform-builder">
```
velkymx: dynaform? We don't use this. We need to use ONLY bootstrap classes.

velkymx: image

PR #103, opened 2019-03-11
Make bootstrap an external/peer dependency

chipit24:
Currently, bootstrap was included in the bundle output. This PR makes it an external dependency so it plays well with the environment where it's used (to avoid conflicts with bootstrap versions). This change is required for the `switch` property on the `b-form-checkbox` component to actually show a switch (required for https://github.com/ProcessMaker/modeler/pull/278).

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,6 @@
   "dependencies": {
     "@fortawesome/fontawesome-free": "^5.6.1",
     "@processmaker/vue-form-elements": "^0.5.0",
-    "bootstrap": "^4.1.3",
     "css-tree": "^1.0.0-alpha.29",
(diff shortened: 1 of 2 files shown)


velkymx:
@chipit24 when I attempt to build this, I get an npm dependency missing.

tdondich:
@chipit24 How do I use this commit, when testing/building/developing locally with a clean node_modules?  I get problems.

chipit24:
@velkymx @tdondich Good catch! `bootstrap` should still be in `devDependencies`. `npm i` no longer installs peer dependencies automatically and I never tested it with a fresh `node_modules`.

PR #121, opened 2019-03-14
Add error validations inspector

### LINKED ISSUE #108: Validation Issues
We need to update the validation to validate the requirements on the inspector. Currently we've added the Field Name to controls that don't need a field name (like columns). This needs to be removed and replaced by proper validation.

caleeli:
Fixes #108  
This PR includes a inspector of validations issues:

ValidationIspector


velkymx:
@caleeli I've merged this to review, but the error drawer should be on the bottom like the modeler

PR #127, opened 2019-03-19
Add  rich text (wysiwyg) editor

### LINKED ISSUE #125: Add Rich Text to Toolbar
I want be able to be able to drag a rich text WYSIWYG block from the controls onto the canvas. 

Use this icon: pencil-ruler
label: Rich Text

chipit24:
Fixes #125.

**Important changes:**
* `eslint-plugin-jest` has been added; run `npm ci` or `npm i` before testing.
* New test added; all tests should pass when running `npm test`.

rich-text

### DIFF
--- a/src/components/vue-form-builder.vue
+++ b/src/components/vue-form-builder.vue
@@ -9,10 +9,10 @@
                                v-model="controls"
                                :options="{sort: false, group: {name: 'controls', pull: 'clone', put: false}}"
                                :clone="cloneControl">
-                        <div class="d-flex align-content-center flex-wrap p-2"
+                        <div class="d-flex align-items-center flex-wrap m-2 mb-3"
                              v-for="(element, index) in controls"
                              :key="index">
-                            <div class="control-icon">
+                            <div class="control-icon d-flex align-items-center">
                                 <img v-if="element['editor-icon']" :src="element['editor-icon']">
                                 <i v-if="element['fa-icon']" :class="element['fa-icon']"></i>
                             </div>
@@ -349,19 +349,11 @@
 
 <style lang="scss" scoped>
     .control-icon {
-        width: 20px;
-        margin-right: 8px;
+        width: 30px;
+        font-size: 20px;
 
         img {
-            max-width: 20px;
-            max-height: 20px;
-        }
-
-        text-align: right;
-
-        i {
-            font-size: 24px;
-            margin-right: 8px;
+          height: 20px;
         }
     }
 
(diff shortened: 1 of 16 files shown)


src/components/vue-form-builder.vue
```diff
@@ -165,6 +165,8 @@
 
   import BootstrapVue from "bootstrap-vue";
 
+  import '@processmaker/vue-form-elements/dist/vue-form-elements.css';
```
chipit24: If having a separate CSS import is preferred, styles could be inlined. See https://cli.vuejs.org/guide/build-targets.html#library:
> * `dist/myLib.css`: Extracted CSS file (can be forced into inlined by setting `css: { extract: false }` in `vue.config.js`)

velkymx:
AH! OK the editor should only show in the edit screen. The preview screen should only show the final resulting HTML

chipit24:
@velkymx I've updated component to work as expected:
rich_text_editor

**Note:** It requires the changes from x to be merged and published; blocked until then.

velkymx:
@chipit24 @milaendo can we add default text to the control. It's confusing if there isn't anything there.

milaendo:
@chipit24 @velkymx After digging into this I have run into an interesting bug. Ill do my best to explain:

in the form-builder-controls.js There is "content" being set in the config for the html editor, which should show on the canvas when the control is dragged on as the default text. However, in the vue-form-elements the value is being set equal to the content. in the config if I add "value: click here to edit", it will show as the default text but I am unable to edit it in both the inspector as well as the canvas.

Im happy to jump on a call if this doesnt make sense.

chipit24:
I've updated the PR.

@milaendo It's possible that this PR wasn't referencing the latest `@processmaker/vue-form-elements` package when you tested it. Please pull down the latest changes, make sure to run `npm i` or `npm ci`, and test again. If it's still failing we can hop on a call or you can post a video of the problem.

milaendo:
@chipit24 That seems to have been the problem! all is fixed :) my only feedback now is that the control icon is a bit larger than all of the others 
Screen Shot 2019-04-01 at 11 06 06 AM

chipit24:
Updated styles:

Screen Shot 2019-04-01 at 3 18 05 PM

velkymx:
@chipit24 please resolve these conflicts to merge.

velkymx:
@milaendo @chipit24 Move the control to the top of the menu and let's merge it.

chipit24:
@velkymx Done.

PR #134, opened 2019-03-28
Add the field name back, just not required

### LINKED ISSUE #132: Name Missing Text
I cannot name my text block anymore. I need to be able to target it for visibility rules and other options, and cant. Please re-add the Field Name

nolanpro:
Fixes #132

Add the field name back, just not required

Kapture 2019-03-28 at 9 28 56

### DIFF
--- a/src/form-builder-controls.js
+++ b/src/form-builder-controls.js
@@ -116,6 +116,15 @@ export default [{
         },
         inspector: [
             {
+                type: "FormInput",
+                field: "name",
+                config: {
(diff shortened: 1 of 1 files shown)


tdondich:
I dont like this. This is only being used for css targeting but it will mess up the data model when in use.

I would prefer we rethink this.  Ideally with a different attribute name.  Let's discuss today.

velkymx:
I've started a list of inspector fields for clarification. We can revisit this after.

PR #140, opened 2019-04-09
Redesign Builder

### LINKED ISSUE #133: New Design
Screenshot 2019-03-27 12 12 56

HTML / CSS

[vue-screen-builder-20190328-AB.txt](https://github.com/ProcessMaker/vue-form-builder/files/3018722/vue-screen-builder-20190328-AB.txt)

tuantruong625:
- Elements are wrapped in a bootstrap `card` class
- `card-header`, `card-body`, `card-footer`  supporting classes are used when possible

**Builder**

Screen Shot 2019-04-18 at 11 24 26 AM

**Preview**

Screen Shot 2019-04-18 at 11 23 42 AM

Fixes #133, #149

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -94,18 +94,16 @@
                   <i class="fas fa-times-circle text-danger mr-3"></i>
                   {{$t('Invalid JSON Data Object')}}
                 </div>
-                <a v-for="(validation,index) in validationErrors" :key="index" href="javascript:void(0)"
-                  class="validation__message d-flex align-items-center p-3"
-                  @click="focusInspector(validation)"
-                >
+                <b-button variant="link" class="validation__message d-flex align-items-center p-3"
+                          v-for="(validation,index) in validationErrors"
+                          :key="index"
+                          @click="focusInspector(validation)">
                   <i class="fas fa-times-circle text-danger d-block mr-3"></i>
-                  <span class="ml-2 text-dark font-weight-bold">
+                  <span class="ml-2 text-dark font-weight-bold text-left">
                     {{ validation.item.component }}
                     <span class="d-block font-weight-normal">{{ validation.message }}</span>
                   </span>
-                </a>
-
-
+                </b-button>
                 <span v-if="!allErrors" class="d-flex justify-content-center align-items-center h-100">No Errors</span>
             </div>
 
--- a/src/App.vue
+++ b/src/App.vue
@@ -49,9 +49,9 @@
                 </div>
             </div>
 
-            <div class="data-container col-6">
-              <div id="data-input" class="h-50 border overflow-auto">
-                  <div class="card-header d-flex align-items-center">
+            <div class="data-container col-6 shadow-sm pl-0">
+              <div id="data-input" class="h-25rem border overflow-auto">
+                  <div class="card-header d-flex align-items-center sticky-top header-bg">
                       Data Input
                       <div class="ml-auto">
                           <span v-if="previewInputValid">
@@ -65,11 +65,12 @@
                       </div>
                   </div>
                   <div class="card-body mb-5">
-                      <form-text-area rows="18" v-model="previewInput"></form-text-area>
+                      <form-text-area rows="8" v-model="previewInput"></form-text-area>
                   </div>
               </div>
-              <div id="data-preview" class="h-50 border overflow-auto mb-5">
-                  <div class="card-header">
+
+              <div id="data-preview" class="h-25rem border overflow-auto">
+                  <div class="card-header sticky-top header-bg">
                       Data Preview
                   </div>
                   <vue-json-pretty :data="previewData" class="card-body"></vue-json-pretty>
@@ -216,5 +217,13 @@
         min-height: 100%;
         max-height: 100%;
         overflow: hidden;
+    }
+
+    .h-25rem {
+      height: 25rem;
+    }
+
+    .header-bg {
+      background: #f7f7f7;
     }
 </style>
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,4 +1,5 @@
 <template>
+  <div class="container h-100">
     <div id="app" class="card h-100">
       <div class="card-header">
         <div class="row">
@@ -124,6 +125,7 @@
             </div>
         </div>
     </div>
+  </div>
 </template>
 
 <script>
--- a/src/App.vue
+++ b/src/App.vue
@@ -26,7 +26,6 @@
             </div>
         </div>
 
-
         <computed-properties v-model="computed" ref="computedProperties"></computed-properties>
         <custom-CSS v-model="customCSS" ref="customCSS" :cssErrors="cssErrors"/>
         <vue-form-builder ref="builder" @change="updateConfig" v-show="displayBuilder"/>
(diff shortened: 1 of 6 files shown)


velkymx:
@tuantruong625 this isn't close enough to the provided HTML.

velkymx:
@tuantruong625 Module Warning (from ./node_modules/eslint-loader/index.js):
error: 'v-model' directives aren't supported on <button> elements (vue/valid-v-model) at src/components/renderer/form-button.vue:3:64:
  1 | <template>
  2 |     <div class="form-group">
> 3 |         <button @click="click" :class="classList" :name="name" v-model="fieldValue">{{ label }}</button>
    |                                                                ^
  4 |     </div>
  5 | </template>

velkymx:
Footer gutter is too large.

Screenshot 2019-04-18 14 23 19

velkymx:
Remove the focus red outline. Only show red if the validation calls for it.

velkymx:
Inspector seems a bit buggy. You can see in this screen the wrong arrow in the accordian is shown for the open element, and I had to click multiple times to open it.  

Screenshot 2019-04-18 14 26 23

velkymx:
This white space between the page edit / delete buttons is too large. Make the select wider to fill it in.
 
Screenshot 2019-04-18 14 27 58

velkymx:
Divider missing between the pages toolbar and the scrolling form content below.

Screenshot 2019-04-18 14 30 03

velkymx:
The buttons look to be different sizes. You can clearly see it when you compare the trash bin icons.

Screenshot 2019-04-18 14 31 19

velkymx:
Scrolling clarification.

Screenshot 2019-04-18 14 35 25

velkymx:
You can't actually "drag an element here" it's above and below - and it shouldn't be. If you drag onto a space (above or below) then it should show that div and allow you to drop the control onto it. If there are no controls on the canvas it should show the drag element here box by default.

chipit24:
> Remove the focus red outline. The grey outline frame is the focus state. The unfocused state should remove the frame. Only show red if the validation calls for it.
> 
> * Normal: No frame
> * Focused: Grey frame with title and trash bin
> * Error: Red frame

@velkymx This is how we interpreted these changes (for some reason when exporting the gif it removed the grey background from the card headers on focus, but it's there):
spark-sb-changes

velkymx:
Can we try it like this? I can get some stakeholder feedback.

src/App.vue
```diff
@@ -120,13 +193,20 @@
       previewInput() {
         if (this.previewInputValid) {
           // Copy data over
+          JSON.stringify(this.previewInput)
```
chipit24: This line doesn't appear to do anything (the result isn't assigned to a variable).

src/App.vue
```diff
@@ -68,16 +134,19 @@
...
+  // Bring in our initial set of controls
   import controlConfig from "./form-builder-controls";
   import globalProperties from "./global-properties";
-
-
   import {
     FormTextArea,
   } from "@processmaker/vue-form-elements";
 
+let Validator = require('validatorjs');
```
chipit24: Is there a reason we're using `require` here instead of `import`?

src/App.vue
```diff
@@ -1,62 +1,128 @@
...
-                <li class="nav-item">
-                    <a class="nav-link" @click="openCustomCSS" href="#">{{$t('Custom CSS')}}</a>
-                </li>
-            </ul>
-        </nav>
+    <div id="app" class="card h-100">
+      <div class="card-header">
+        <div class="row">
+          <div class="col">
+            <div class="btn-group btn-group-sm" role="group" aria-label="Basic example">
```
chipit24: This whole `btn-group` block can be replaced with:

```html
<b-button-group size="sm">
  <b-button :variant="displayBuilder? 'outline-secondary' : 'secondary'" @click="mode = 'editor'">{{ $t('Design') }}</b-button>
  <b-button :variant="!displayBuilder? 'outline-secondary' : 'secondary'" @click="mode = 'preview'">{{ $t('Preview') }}</b-button>
</b-button-group>
```

This uses the available `bootstrap-vue` components instead of classes. Updating the style based on `variant` means the `isBuilderActive` and `isPreviewActive` computed properties can be removed; these computed properties were named in such a way that it seemed like they should return booleans.

The current `aria-label` provides no additional value.

We can use translation! 🎉

Since this is a toggle, removing `isPreviewActive` and using `isBuilderActive` and `!isBuilderActive` may simply be enough.

src/App.vue
```diff
@@ -1,62 +1,128 @@
...
-            <div id="data-preview" class="w-25 border overflow-auto mb-5">
-                <div class="card-header">
-                    {{$t('Data Preview')}}
+            <div v-if="showValidationErrors" class="validation-panel position-absolute shadow border overflow-auto" :class="{'d-block':showValidationErrors && validationErrors.length}">
+                <div v-if="!previewInputValid" class="p-3 font-weight-bold text-dark">
+                  <i class="fas fa-times-circle text-danger mr-3"></i>
+                  {{$t('Invalid JSON Data Object')}}
                 </div>
-                <vue-json-pretty :data="previewData" class="card-body"></vue-json-pretty>
+                <a v-for="(validation,index) in validationErrors" :key="index" href="javascript:void(0)"
```
chipit24: I'm not a fan of using `href="javascript:void(0)"` to disguise a link as a button. I think it might be better to use `<b-button variant="link">Link</b-button>`.

src/App.vue
```diff
@@ -1,62 +1,128 @@
...
+                    {{ validation.item.component }}
+                    <span class="d-block font-weight-normal">{{ validation.message }}</span>
+                  </span>
+                </a>
+
+
+                <span v-if="!allErrors" class="d-flex justify-content-center align-items-center h-100">No Errors</span>
+            </div>
+
+            <div class="ml-3" @click="showValidationErrors =! showValidationErrors">
```
chipit24: `=!` looks like a typo; it should be `!==`.

src/App.vue
```diff
@@ -192,4 +312,23 @@
...
+      width: 21.35rem;
+      bottom: 3.5rem;
+      right: 0;
+    }
+    .dataInput {
+      margin-top: -25px;
+    }
+
+    .h-80 {
+      height: 80vh;
```
chipit24: This might conflict with the bootstrap util classes.

velkymx:
@tuantruong625 getting closer. Here is a side-by-side of the provided HTML and the current implementation. I've highlighted some of the discrepancies.  

sidebyside

src/App.vue
velkymx: I agree

src/App.vue
velkymx: We should try and eliminate custom CSS as much as possible. We should be using only bootstrap classes.

velkymx:
Placeholder text isn't being rendered in the design mode. 

Screenshot 2019-04-22 15 18 04

tuantruong625:
> Placeholder text isn't being rendered in the design mode.
> 
> Screenshot 2019-04-22 15 18 04



> Placeholder text isn't being rendered in the design mode.
> 
> Screenshot 2019-04-22 15 18 04

This bug appears on develop as well.

Issue #162, opened 2019-05-07
Design Broken


velkymx:
When I load the screen builder into the Spark project it looks terrible. We need to ALWAYS test with the tool inside of Spark.

![Screenshot 2019-05-07 10 04 51](https://user-images.githubusercontent.com/535300/57318629-9ef5b000-70af-11e9-8d25-d9616ee769c1.png)

velkymx:
![Screenshot 2019-05-07 10 06 50](https://user-images.githubusercontent.com/535300/57318725-d6645c80-70af-11e9-852b-71fe333515b7.png)

velkymx:
![Screenshot 2019-05-07 10 04 51](https://user-images.githubusercontent.com/535300/57318831-10cdf980-70b0-11e9-9106-f8b1b86ae6b5.png)

velkymx:
1. The preview inspector is square instead of rounded.
1. The footer is off screen instead of in the footer
1. The preview form isn't scrolling - the whole page is.
1. Custom CSS selectors aren't working properly
1. Rich Text doesn't load
1. ProcessMaker theme is being overwritten with the wrong fonts and colors

velkymx:
@tuantruong625 @chipit24 @ssshake See my comments.

PR #164, opened 2019-05-09
Design Bugs

tuantruong625:
Reference #162.

Screen Shot 2019-05-09 at 3 26 38 PM

To fix button sizes in header, had to use custom css.

Fixes some of the bigger rendering issues: General layout/size/padding issuyes, rich text editor rendering, etc. Rest is still WIP.


src/App.vue
```diff
@@ -48,9 +47,9 @@
                                           v-on:css-errors="cssErrors = $event"/>
                   </div>
               </div>
-            </div>
+            </b-col>
 
-            <div class="data-container col-4 border overflow-auto pr-0 pl-0">
+            <b-col cols="4" class="data-container border overflow-auto pr-0 pl-0 rounded">
```
velkymx: This should only need the card wrapper! Why are we adding all of these classes???

velkymx: https://getbootstrap.com/docs/4.3/components/card/

velkymx:
Here's the HTML from the New Design ticket.

velkymx:
Getting an error when I do npm run serve

```
Module Warning (from ./node_modules/eslint-loader/index.js):
error: 'v-model' directives aren't supported on <button> elements (vue/valid-v-model) at src/components/renderer/form-button.vue:3:64:
  1 | <template>
  2 |     <div class="form-group">
> 3 |         <button @click="click" :class="classList" :name="name" v-model="fieldValue">{{ label }}</button>
    |                                                                ^
  4 |     </div>
  5 | </template>
  6 | 
```

ssshake:
button @click="click" :class="classList" :name="name" v-model="fieldValue"

This was a pre-existing issue and does not break the app, just throws a warning. Prior discussions led to leaving this as is in this PR as it's out of scope and then fixing it outside of this PR.

I've split this out into a separate issue: https://github.com/ProcessMaker/spark-screen-builder/issues/175

caleeli:
Should we have to do any additional change in Spark, I see that this style is still wrong:

The border of the page selector is still square

image

Screenshot 2019-05-07 10 06 50

velkymx:
@caleeli don't worry about that style on the dropdown for now.

PR #197, opened 2019-05-22
Rename “Field Name” to “Variable Name”

### LINKED ISSUE #156: Inspector: Field Name
Lots of confusion around field name label in the sidebar. Change it to be "Variable Name"

### LINKED ISSUE #190: Radio Options Labels Wrong
When trying the add a new option to a Radio control, the popup form reads Field Name and Field Label, when it should be instead something like Value and Content.

chipit24:
Fixes #156, fixes #190.

Screenshots of fix:

Screen Shot 2019-05-22 at 10 40 54 AM

Screen Shot 2019-05-22 at 1 29 22 PM

### DIFF
--- a/src/components/inspector/options-list.vue
+++ b/src/components/inspector/options-list.vue
@@ -51,7 +51,7 @@
             <b-btn v-b-modal.addOptionModal>{{ $t('Add Option') }}</b-btn>
 
             <b-modal centered @cancel="resetAdd" @ok="addNewOption" id="addOptionModal" :title="$t('Add New Option')">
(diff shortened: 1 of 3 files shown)


tdondich:
In the realm of selects that modal should show it as variable value

chipit24:
> In the realm of selects

@tdondich What do you mean by this? You're referring to the modals for options / fields lists (Select, Radio Group, Record List)?

If I rename "Value" to "Variable Value" in the modal, should I rename it in the options list section as well (may not fit well, so I'm assuming no)?
Screen Shot 2019-05-22 at 3 19 36 PM

Issue #171, opened 2019-05-15
Preview Inspector elements should use up available vertical realestate


ssshake:
![image](https://user-images.githubusercontent.com/1675868/57803144-b115c500-7725-11e9-8c85-1fee1c89d1e6.png)

tuantruong625:
- Not as simple as setting `h-50` on data input and preview because of the headers
- setting the height of the text-area dynamically without modifying vue-form-elements repo 

@velkymx Will continue on Tuesday

<img width="1674" alt="Screen Shot 2019-05-17 at 6 09 00 PM" src="https://user-images.githubusercontent.com/26545455/57958910-7fd8f880-78cf-11e9-8aa1-b065358eb98d.png">

tuantruong625:
@velkymx 

The transitions on the accordions are broken because to fill the remaining space when collapsed / close with the data input & preview we set the accordions to a flex container. Which breaks the transitions because it relies on the height of the container to make the transition.

**Video with broken transition**
https://drive.google.com/open?id=1fsWA5OLfUPj-4E-jfCC2VFTs_US4NWFM

To fix the transitions on the accordions will require more work with javascript.

The comprise is setting a fixed height on the data input / preview containers. The only issue is when there is an overflow with the data input there will be double scroll bars. As shown in the screenshot.

<img width="1680" alt="Screen Shot 2019-05-21 at 4 09 05 PM" src="https://user-images.githubusercontent.com/26545455/58131797-947a0100-7bed-11e9-90c0-2b197bc2766a.png">

Do we want to invest more time to fix the animations ?

**Current Implementation**
Or have the data input / preview accordion to behave like the script editor accordions ?

PR #198, opened 2019-05-22
Preview Inspector UI

### LINKED ISSUE #172: Preview Inspector is not the same width as Design Inspector
These should both use the same classes 

image

tuantruong625:
- Preview Inspector uses the full height of the card
- Data input / preview take 50% of the inspector container

Screen Shot 2019-05-22 at 1 29 17 PM

https://drive.google.com/open?id=1XECBNU_X86rXZcLReKm7eHUL5bn73QrH

Fixes #172 , #171

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -356,8 +356,6 @@ import Validator from "validatorjs";
       max-width: 265px;
     }
 
-
-
     .data-collapse {
       height: 300px;
     }
--- a/src/App.vue
+++ b/src/App.vue
@@ -39,7 +39,7 @@
 
         <!-- Preview -->
         <b-row class="h-100" id="preview" v-show="displayPreview">
-          <b-col cols="8" class="overflow-auto h-100 border rounded">
+          <b-col class="overflow-auto h-100 border rounded mr-4">
             <vue-form-renderer ref="renderer"
               v-model="previewData"
               class="p-3 overflow-auto"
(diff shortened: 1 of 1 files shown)


ssshake:
Through discussion today we agreed to role with a fixed height, even though it may cause overflow scroll on smaller screens. Attempting to fix this with flex was causing unpleasant behaviour to the expand/collapse animation.

src/App.vue
```diff
@@ -337,10 +344,21 @@ import Validator from "validatorjs";
...
 
     .list-group-item:last-child {
       border-bottom: 1px solid #dfdfdf !important;
     }
+
+    .preview-inspector {
+      max-width: 265px;
+    }
+
+
```
chipit24: Remove the extra spaces here.

src/App.vue
```diff
@@ -337,10 +344,21 @@ import Validator from "validatorjs";
     }
 
     .data-height {
```
chipit24: This class doesn't seem to be used anymore; can be removed.

src/App.vue
```diff
@@ -50,36 +50,38 @@
...
-                <b-collapse id="dataPreview" visible class="mt-2 overflow-auto">
-                  <vue-json-pretty  :data="previewData" class="p-2 data-height"></vue-json-pretty>
-                </b-collapse>
+              <b-card-body class="p-0 overflow-auto">
+                  <b-button variant="outline"
+                    class="text-left card-header d-flex align-items-center w-100 shadow-none"
+                    @click="showDataInput = !showDataInput">
+                    <i class="fas fa-file-import mr-2"></i>
+                      {{ $t('Data Input') }}
+                    <i class="fas ml-auto" :class="showDataInput ? 'fa-angle-right' : 'fa-angle-down'"></i>
```
chipit24: This should be reversed. Right angle should be visible when card is collapsed, and down angle when card is open–exactly like the Configuration card in the Design page.

src/App.vue
```diff
@@ -50,36 +50,38 @@
...
+                  </b-button>
+
+                  <b-collapse v-model="showDataInput" id="showDataInput">
+                    <form-text-area class="data-input mb-0 data-collapse" v-model="previewInput"></form-text-area>
```
chipit24: Is seems this is required because of the margin on the label for the rendered `form-text-area`. You can clean this up (remove `data-input` and replace it with `h-100`) if you set `display: none` on the label inside this component.

chipit24:
Make sure to port PR changes here to the spark PR as well :)

PR #225, opened 2019-06-05
Adjust draggable options for nested lists

### LINKED ISSUE #221: Placed Elements not available in multi-column
Cannot drag an existing element into a multi-column "Table". I can drag a new Image Control from the sidebar into a table, but once the image is on the form, it cannot be moved into a Table.

chipit24:
Fixes #221.

Video of fix: https://www.dropbox.com/s/0rynwuzcsiq345x/dragging_improved.mov?dl=0.

The functionality for dragging elements into/out of nested draggable lists (table / multi-column) was already correctly implemented. However, due to styling changes, the relative sizes of elements, and the fact that they can change, made it difficult to line up the drop / swap targets to allow dragging into a nested list.

This PR modifies the properties of the draggable lists to make it easier to interact with nested lists. The descriptions of these options is listed here: https://github.com/SortableJS/Sortable#options.

The [SortableJS example](http://sortablejs.github.io/Sortable/#nested) mentions the following, which has been implemented in this PR:
> NOTE: When using nested Sortables with animation, it is recommended that the fallbackOnBody option is set to true. 
It is also always recommended that either the invertSwap option is set to true, or the swapThreshold option is lower than the default value of 1 (eg 0.65).

To make dragging / dropping as smooth as possible, it would be ideal to keep the dragging elements small, the drop targets large, and the elements in the same relative position for nested lists. This is especially clear when there's an empty multi-column table, and you highlight a form element (having the card with the delete button increases it's size significantly).

### DIFF
--- a/src/components/editor/multi-column.vue
+++ b/src/components/editor/multi-column.vue
@@ -5,7 +5,12 @@
                 <template v-for="(item, index) in items">
                     <draggable :class="classColumn(index)"
                             v-model="items[index]"
-                            :options="{group: {name: 'controls'}}"
+                            :options="{
+                              invertSwap: true,
+                              group: {name: 'controls'},
+                              direction: 'horizontal',
+                              emptyInsertThreshold: 5
+                            }"
                             :key="index">
                         <div class="control-item"
                                 :class="{selected: selected === element}"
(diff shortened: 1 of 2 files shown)


chipit24:
@tuantruong625 Update PR.

A few points:
 * Trying to move an element from above the table into the table will be difficult as it will cause the table to jump up and the element to be placed after it, unless the element you are dragging has a small height or the table has a large height.
* The reduced margin on the table element makes it easier to drag items into it; if the margin it too large, draggable might trigger a swap with the entire table before it recognizes the draggable element inside the table.

PR #241, opened 2019-06-17
Fix undefined error message

### LINKED ISSUE #186: Error Message: The undefined field is required.
Screenshot 2019-05-16 17 17 10

Should say the Name Field is required as it does in the error panel

tuantruong625:
- Fixed error messages that rendered `The undefined field is required`
- Change helper texts that reference `data`

**Spark video demo**
https://drive.google.com/open?id=1y3ZYWu_C-5VJDe7HxVCSYgAGHBl7RYds

Fixes #186

### DIFF
--- a/src/form-builder-controls.js
+++ b/src/form-builder-controls.js
@@ -867,7 +867,7 @@ export default [
                 field: "name",
                 config: {
                     label: "Variable Name",
-                    helper: "The name of the image"
+                    helper: "The variable name of the image"
                 }
             },
             {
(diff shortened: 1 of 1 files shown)


chipit24:
@tuantruong625 Do you think image should use "variable name" instead of just "name" as well? I think it makes sense to keep it consistent.

Screen Shot 2019-06-17 at 4 16 59 PM

tuantruong625:
> @tuantruong625 Do you think image should use "variable name" instead of just "name" as well? I think it makes sense to keep it consistent.
> 

@chipit24 Good idea, I will make those changes.

PR #242, opened 2019-06-18
Add drag here placeholder

### LINKED ISSUE #170: "Drag Here" jumps around
You should be able to drag on top of drag here. So change the way we're doing this entirely. Currently drag here is a draggable item itself.

https://www.dropbox.com/s/5l7h7n956ra65r7/drag-here-weirdness.mov?dl=0

tuantruong625:
Screen Shot 2019-06-19 at 1 43 31 PM

**Spark Video Demo**
https://drive.google.com/open?id=1Oxl45COxLT2h--hCeY-x-7EXii09uPTT

Fixes #170

### DIFF
--- a/src/components/vue-form-builder.vue
+++ b/src/components/vue-form-builder.vue
@@ -36,7 +36,7 @@
     </b-col>
 
     <!-- Renderer -->
-    <b-col class="overflow-auto mh-100 pl-4 pr-4 d-flex flex-column">
+    <b-col class="overflow-auto mh-100 pl-4 pr-4 d-flex flex-column position-relative">
       <b-input-group size="sm" class="sticky-top bg-white">
         <b-form-select v-model="currentPage" class="form-control">
           <option v-for="(data, page) in config" :key="page" :value="page">{{ data.name }}</option>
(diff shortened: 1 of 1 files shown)


src/components/vue-form-builder.vue
```diff
@@ -314,6 +319,11 @@ export default {
           .toLowerCase()
           .includes(this.filterQuery.toLowerCase());
       });
+    },
+    containsFormElements() {
```
chipit24: I wonder if this should only check the current page, instead of all pages? If I have multiple empty pages and drag one element onto just one of the empty pages, the "drag an element here" placeholder no longer appears for any of the other empty pages.

src/components/vue-form-builder.vue
```diff
@@ -67,6 +67,11 @@
 
         <hr class="w-100">
       </b-input-group>
+
+        <b-card-body class="d-flex justify-content-center align-items-center drag-placeholder text-center position-absolute" v-if="!containsFormElements">
```
chipit24: Nitpick: This is a `b-card-body` component but it's not inside of a card. I removed the `card-body` class and it still looks the same, so I think it's better to use a regular div here.

Also, this block is indented one more than the blocks around it–it should in on the same line.

src/components/vue-form-builder.vue
```diff
@@ -36,7 +36,7 @@
     </b-col>
 
     <!-- Renderer -->
-    <b-col class="overflow-auto mh-100 pl-4 pr-4 d-flex flex-column">
+    <b-col class="overflow-auto mh-100 pl-4 pr-4 d-flex flex-column position-relative">
```
chipit24: If you remove `pl-4 pr-4` and replace it with `ml-4 mr-4 p-0`, you can use `w-100` on the "drag an element here" container instead of the custom `width: calc(100% - 48px)`.

PR #255, opened 2019-06-24
Adjust mask to cover 100% of element

### LINKED ISSUE #254: Mask doesn't properly cover form-builder element
When an element is selected in design mode, users should not be able to interact with the element as if it was rendered (like in preview). Currently, a mask is used to prevent interaction, but it does not completely cover the element:

Screen Shot 2019-06-24 at 9 56 39 AM

Screen Shot 2019-06-24 at 9 56 30 AM

This issue was noticed here: https://github.com/ProcessMaker/processmaker/pull/2028#pullrequestreview-251927902.

chipit24:
Fixes #254.

Gif of fix (with file-upload component):
working_mask

Screenshots of fix:
Screen Shot 2019-06-24 at 10 30 15 AM

Screen Shot 2019-06-24 at 10 29 36 AM

Screen Shot 2019-06-24 at 10 29 18 AM

### DIFF
--- a/src/components/vue-form-builder.vue
+++ b/src/components/vue-form-builder.vue
@@ -124,15 +124,14 @@
               </button>
             </div>
 
-            <div class="card-body m-0 pb-4 pt-4 position-relative">
-              <component
-                :class="elementCssClass(element)"
-                v-bind="element.config"
-                :is="element['editor-component']"
-                @input="element.config.interactive ? element.config.content = $event : null"
-              />
-              <div v-if="!element.config.interactive" class="mask"></div>
-            </div>
+            <component
+              tabindex="-1"
+              class="prevent-interaction card-body m-0 pb-4 pt-4"
+              :class="elementCssClass(element)"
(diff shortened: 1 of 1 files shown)


src/components/vue-form-builder.vue
```diff
@@ -469,15 +469,10 @@ $header-bg: #f7f7f7;
   .mask {
```
tuantruong625: Could we use the bootstrap helper classes for width, height, and position?

chipit24: I thought about this; we could use `w-100 h-100 position-absolute`, but I didn't see any helpers for `top` and `left` positioning. I wanted to keep the `mask` class since it gives the div meaning, and makes it easier to find when debugging.

I wonder if we can use something like `pointer-events: none` instead of a mask though; will look into it and update PR.

tuantruong625: I like `pointer-events: none` instead of defining what percent of the mask should cover.

PR #263, opened 2019-06-25
Replace config accordion with variable and design

chipit24:
Fixes #249.

This PR removes the Configuration accordion, and replaces it with Variable and Design accordions. The Variable accordion contains a specified list of form types that correspond to variable data, while the Design accordion contains all the other (design) fields.

Video of fix: https://www.dropbox.com/s/s256p0rvucvaz69/var_reorg_working.mov?dl=0.

PR #274, opened 2019-06-28
Variable Data Format Enhancements

marcoAntonioNina:
- Enable UI Variable
- enable inspector

https://github.com/ProcessMaker/processmaker/issues/2001

### DIFF
--- a/src/components/vue-form-builder.vue
+++ b/src/components/vue-form-builder.vue
@@ -159,6 +159,7 @@
             variant="outline"
             class="text-left card-header d-flex align-items-center w-100 outline-0 text-capitalize shadow-none"
             @click="showVariable = !showVariable"
+            v-if="hasVariableConfig(inspection.inspector)"
           >
             <i class="fas fa-cog mr-2"/>
             {{ $t('Variable') }}
@@ -168,7 +169,7 @@
             />
           </b-button>
 
-          <b-collapse id="variableConfig" class="mt-2">
+          <b-collapse id="variableConfig" class="mt-2" v-if="hasVariableConfig(inspection.inspector)">
             <template v-for="(item, index) in inspection.inspector">
               <template v-if="item.panel && item.panel === 'variable'">
                 <component
@@ -399,6 +400,9 @@ export default {
     },
   },
   methods: {
+    hasVariableConfig(inspector) {
+      return inspector && inspector.filter(item => item.panel==='variable').length > 0;
+    },
     updateConfig(items) {
       this.config[this.currentPage].items = items;
     },
--- a/src/components/vue-form-builder.vue
+++ b/src/components/vue-form-builder.vue
@@ -139,34 +139,60 @@
     <!-- Inspector -->
     <b-col class="overflow-hidden h-100 p-0 inspector-column">
       <b-card no-body class="p-0 h-100">
-        <b-card-header>{{ $t('Inspector') }}</b-card-header>
-
+        <b-card-header>
+          {{ $t('Inspector') }}
+        </b-card-header>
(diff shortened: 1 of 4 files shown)


src/form-builder-controls.js
```diff
@@ -61,41 +62,171 @@ const colorProperty = {
...
+  type: 'FormInput',
+  field: 'validation',
+  panel: 'variable',
+  config: {
+    label: 'Validation',
+    name: 'validation',
+    helper: 'Validation rules ensure the integrity and validity of the data.',
+  },
+};
+const ReadOnly = {
```
chipit24: I like that this was broken out into a reusable object 👍 

@velkymx @marcoAntonioNina

There is already a PR open for adding the "read only" property: https://github.com/ProcessMaker/screen-builder/pull/265/files. Not sure why this work was duplicated here.

This also seems to set the `disabled` attribute, not the `readonly` attribute. Bootstrap components accept `readonly`: https://getbootstrap.com/docs/4.0/components/forms/#readonly. Was this intentional, or do we actually want this to disable the component? From the perspective of an average user, the difference may to be very noticeable, and `readonly` only seems to work on text input and textarea.

src/form-builder-controls.js
```diff
@@ -131,15 +262,18 @@ export default [
...
+            validation: 'required',
+            options: [
+              {
+                value: 'normal',
+                content: 'Normal',
+              },
+              {
+                value: 'bold',
+                content: 'Bold',
+              },
```
chipit24: If formatting changes are required, they should be added to the `.eslintrc.js` file. A good bit of this PR is unnecessary formatting changes; please have your code editor format based on the ESLint config file.

src/components/vue-form-builder.vue
```diff
@@ -149,68 +149,104 @@
...
       <b-card no-body class="p-0 h-100">
-        <b-card-header>{{ $t('Inspector') }}</b-card-header>
-
+        <b-card-header>
+          {{ $t('Inspector') }}
+        </b-card-header>
         <b-card-body class="p-0 h-100 overflow-auto">
+
           <b-button
-            v-b-toggle.configuration
```
chipit24: There is already a PR for this change that just needs the copy updated: https://github.com/ProcessMaker/screen-builder/pull/263. And the implementation here doesn't seem to work as expected (the variable accordion is actually in the "open" state but empty):

Screen Shot 2019-07-02 at 1 51 04 PM

caleeli: This requirement comes from this ticket:
https://github.com/ProcessMaker/processmaker/issues/2001

In which requires the read only property must be inside the Variable panel:
image

src/form-builder-controls.js
caleeli: Following the requirement in the ticket: https://github.com/ProcessMaker/processmaker/issues/2001

The readonly property must be part of the Variable panel

image

src/components/vue-form-builder.vue
chipit24: I assume there was a mixup and @marcoAntonioNina grabbed https://github.com/ProcessMaker/processmaker/issues/2001 and started doing work in this repo, even though there have already been tickets and PRs created for that issue in this repo.

chipit24:
PRs for adding the "Variable" and "Design" accordions, re-naming "Variable" to "Key Name", and adding the `readonly` prop have been merged. The only thing left that this PR should handle is adding the "Data Type" dropdown:

Screen Shot 2019-07-03 at 1 55 55 PM

caleeli:
Please review #288 instead

PR #276, opened 2019-07-02
Fix calc field undefined error.

### LINKED ISSUE #246: Calc Fields Undefined Error
If you have calculated fields on your form, when you open your form for the first time, you will get errors fo "undefined" because the values are not available to make the calculations. These errors should not show up in this case.

tuantruong625:
**How to test**
- Click Calc Button
- Click `Add property` button on modal

**Expected**
No console errors

Adding `typescript ` to the list of languages is required because typescript and javascript have distinct colourizers, but share the same implementation for the language smarts .

**Spark Video Demo**
https://drive.google.com/file/d/1e0twQBNRaY1NFUjaUHQ90ggVgZCmTZZP/view

Fixes #246

### DIFF
--- a/src/components/computed-properties.vue
+++ b/src/components/computed-properties.vue
@@ -104,7 +104,7 @@
         </div>
         <textarea v-show="!isJS" name="formula" v-model="add.formula" class="form-control editor" :class="{'is-invalid':!add.formula}"/>
         <div v-show="isJS" class="editor-border" :class="{'is-invalid':!add.formula}"/>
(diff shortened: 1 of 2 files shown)


chipit24:
To add some more context. From https://github.com/microsoft/monaco-editor-webpack-plugin/tree/1d1473de8ed44d999cb1f6515ce00a4fa67e87ab#options:

> Some languages share the same web worker. If one of the following languages is included, you must also include the language responsible for instantiating their shared worker:
> 
> | Language      | Instantiator  |
> | ------------- | ------------- |
> | javascript    | typescript    |
> | handlebars    | html          |
> | scss, less    | css           |

PR #288, opened 2019-07-10
Add support for Variable Data Types

### LINKED ISSUE #287: Variable Panel
Add support for Variable Types and reorganize the variable panel with the following structure:

image

caleeli:
Resolves #287 

This PR includes:

- Add data type property in Variable Panel
- Refresh form in Design view always (refresh data types validations)
-  Add validation error on Multiselect control (used to select Data Type that is a required property).
- When control is read only, it disable the control so end user can not change the value.

image

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -198,8 +198,10 @@ export default {
     MonacoEditor,
   },
   watch: {
-    mode() {
-      this.previewData = this.previewInput ? JSON.parse(this.previewInput) : null;
+    mode(mode) {
+      if (mode === 'preview') {
+        this.previewData = this.previewInput ? JSON.parse(this.previewInput) : null;
+      }
     },
     config() {
       // Reset the preview data with clean object to start
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,6 +1,6 @@
 <template>
   <b-container class="h-100" id="screen-builder-container">
-    <b-card no-body class="h-100 bg-white" id="app">
+    <b-card no-body class="h-100 bg-white border-top-0" id="app">
       <!-- Card Header -->
       <b-card-header>
         <b-row>
@@ -111,7 +111,7 @@
           </button>
         </div>
 
-        <div v-if="showValidationErrors" class="validation-panel position-absolute shadow border overflow-auto" :class="{'d-block':showValidationErrors && validationErrors.length}">
+        <div v-if="showValidationErrors" class="validation-panel position-absolute border-left border-top overflow-auto" :class="{'d-block':showValidationErrors && validationErrors.length}">
           <div v-if="!previewInputValid" class="p-3 font-weight-bold text-dark">
             <i class="fas fa-times-circle text-danger mr-3"/>
             {{ $t('Invalid JSON Data Object') }}
@@ -317,6 +317,15 @@ export default {
 <style lang="scss">
     @import "~bootstrap/dist/css/bootstrap";
 
+    $validation-panel-bottom: 3.5rem;
+    $validation-panel-right: 0;
+    $validation-panel-height: 10rem;
+    $validation-panel-width: 21.35rem;
+    $primary-white: #f7f7f7;
+
+    $preview-inspector-width: 265px;
+    $data-collapse-height: 300px;
+
     html,
     body {
         height: 100%;
@@ -326,15 +335,15 @@ export default {
     }
 
     .header-bg {
-      background: #f7f7f7;
+      background: $primary-white;
     }
 
     .validation-panel {
-      background: #f7f7f7;
-      height: 10rem;
-      width: 21.35rem;
-      bottom: 4rem;
-      right: 0;
+      background: $primary-white;
+      height: $validation-panel-height;
+      width: $validation-panel-width;
+      bottom: $validation-panel-bottom;
+      right: $validation-panel-right;
     }
 
     .card-header {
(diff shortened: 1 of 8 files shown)


src/App.vue
```diff
@@ -198,10 +198,8 @@ export default {
     MonacoEditor,
   },
   watch: {
-    mode(mode) {
-      if (mode === 'preview') {
-        this.previewData = this.previewInput ? JSON.parse(this.previewInput) : null;
-      }
+    mode() {
+      this.previewData = this.previewInput ? JSON.parse(this.previewInput) : null;
```
chipit24: Just curious, what was the reason for this change? I don't seem to get any validation errors when setting the data type. For example, I drag a line input, change it's type to `Integer`, and have no errors or issues submitting string data with the form.

src/VariableDataTypeProperties.js
```diff
@@ -0,0 +1,104 @@
+export const DataTypeProperty = {
```
chipit24: I think this file can be DRY'ed up. Something like:

```js
const string = { value: 'string', content: 'Text' };
const int = { value: 'int', content: 'Integer' };
const float = { value: 'float', content: 'Decimal' };
const datetime = { value: 'datetime', content: 'Datetime' };
const date = { value: 'date', content: 'Date' };
const currency = { value: 'currency', content: 'Currency' };
const boolean = { value: 'boolean', content: 'Boolean' };

const allOptions = [string, int, float, datetime, date, currency, boolean];
const allOptionsWithoutDate = allOptions.filter(option => ![date, datetime].includes(option));

function dataTypeFactory(options) {
  return {
    type: 'FormMultiselect',
    field: 'dataFormat',
    panel: 'variable',
    config: {
      label: 'Data Type',
      name: 'Data Type',
      helper: 'The data type specifies what kind of data is stored in the variable.',
      validation: 'required',
      options,
    },
  };
}

export const DataTypeProperty = dataTypeFactory(allOptions);
export const DataTypeWithoutDateProperty = dataTypeFactory(allOptionsWithoutDate);
export const DataTypeBooleanProperty = dataTypeFactory([boolean]);
export const DataTypeDateTimeProperty = dataTypeFactory([date, datetime]);
```

src/components/vue-form-renderer.vue
```diff
@@ -30,7 +30,7 @@
           :name="element.config.name !== undefined ? element.config.name : null"
           v-bind="element.config"
           :is="element.component"
-          :disabled="element.config.interactive"
+          :disabled="element.config.readonly || element.config.interactive"
```
chipit24: I'm not sure we want to mark `readonly` inputs as `disabled`. If an input is disabled, a user cannot interact with it at all, and **the input is not submitted**. This behaviour could be unexpected. If we want to actually disable the input, then we should change the wording in the inspector to reflect that.

MDN describes the readonly attribute here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#readonly.

src/components/inspector/form-multiselect.vue
```diff
@@ -61,6 +71,27 @@ export default {
...
     },
+    value: {
+      immediate: true,
+      handler() {
+        this.selected = null;
+        // Load selected item.
+        if (this.value) {
+          this.options.forEach(item => {
+            if (item && item.value && item.value === this.value) {
+              this.selected = item;
```
chipit24: If you refactor the `multiselect` component to use `custom-label`, like in the `page-select.vue` file, the two watchers can be removed. Like:

```
<multiselect
  v-bind="$attrs"
  v-on="$listeners"
  ...
  :options="options.map(option => option.value)"
  :custom-label="getLabelFromValue"
>
...
  ...
  inheritAttrs: false,
  ...
  methods: {
    getLabelFromValue(value) {
      const selectedOption = this.options.find(option => option.value == value);
      return selectedOption ? selectedOption.content : null;
    },
  },
  ...
```

src/App.vue
caleeli: Some times the data type validation is not updated in preview mode

Steps to reproduce:

1. Add and input with Integer data type
2. Go to preview
3. Put a valid integer (it works), put an invalid integer (and show the validation feedback)
4. Go to design mode
5. Change the Data Type to Decimal
6. Go to preview
7. Put an valid decimal but the validation fails, put an integer passes
8. Finally when you change to design and then back to preview, the validation works as expected

Debugging I see that some times (in step 6)  when we back to preview mode, the v-model for this controls are not properly bind to previewData, keeping it value to undefined even if the user changes the value in the input control.

Then I see that if always setup the `previewData` when the mode changes, this issue doesn't happens.

caleeli: Ok I found the issue, in the vue-form-elements PR, caused by modifying directly the transientData inside a watcher. Now it was fixed.

src/components/vue-form-renderer.vue
caleeli: You are right, this change was removed. But actually there is a problem with controls like select, checkbox, etc. where the readonly prop does not block the end user to change the value of the control.

src/components/inspector/form-multiselect.vue
caleeli: done

src/VariableDataTypeProperties.js
caleeli: done

caleeli:
Please review this PR linked to this PR for the vue-form-elements: 
https://github.com/ProcessMaker/vue-form-elements/pull/54
(branch: feature/53)

PR #340, opened 2019-08-12
Change frame styling in tables

### LINKED ISSUE #327: Element Frame Inconsistant
The selected element frame is different when it's inside of a multi-column vs outside. 

Inside:

image

Outside:

image

tuantruong625:
Selected state for elements inside multi-column are consistent with the other states.    

Screen Shot 2019-08-12 at 12 25 21 PM

https://drive.google.com/open?id=1DyhdKg3_QXhQAREtvnjkEvTT1zfRklix

Fixes #327

### DIFF
--- a/src/components/editor/multi-column.vue
+++ b/src/components/editor/multi-column.vue
@@ -18,16 +18,24 @@
               @click.stop="inspect(element)"
             >
               <div v-if="element.container" @click.stop="inspect(element)">
-                <div class="m-2">
-                  <button
+                <div class="m-2 card border-0">
+                  <div
                     v-if="selected === element"
-                    class="element-delete-btn btn btn-sm btn-danger-outline ml-auto position-absolute"
-                    @click="deleteItem(index, row)"
+                    class="card-header form-element-header d-flex align-items-center border rounded"
                   >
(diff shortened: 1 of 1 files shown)


src/components/editor/multi-column.vue
```diff
@@ -18,16 +18,24 @@
...
+                    class="card-header form-element-header d-flex align-items-center border rounded"
                   >
-                    <i class="far fa-trash-alt text-danger"/>
-                  </button>
+                    <i class="fas fa-arrows-alt-v mr-1 text-muted"/>
+                    <i v-if="element.config.icon" :class="element.config.icon" class="mr-2 ml-1"/>
+                    {{ element.config.name || $t('Key Name') }}
+                    <button
+                      class="btn btn-sm btn-danger ml-auto"
+                      @click="deleteItem(index)"
```
chipit24: You should continue to pass `row` to `deleteItem`:
```
@click="deleteItem(index, row)"
```
The 2nd argument is passed as the `start` parameter to `splice`, and is required. Omitting it seems to cause the delete function to just remove the 1st item in the column, even if a user clicked to delete another item.

PR #342, opened 2019-08-12
Validate page names are required and unique

### LINKED ISSUE #304: Unique Page Names
Screen editor page names should be unique identifiers. Currently you can name the pages on one screen the same. There should be some validation for unique name on page tabs when one is created or existing one is edited.

nolanpro:
Fixes #304

### DIFF
--- a/src/components/computed-properties.vue
+++ b/src/components/computed-properties.vue
@@ -78,7 +78,7 @@
         </template>
       </b-table>
       <template slot="modal-footer">
-        <span></span>
+        <span />
       </template>
     </template>
 
--- a/src/components/vue-form-builder.vue
+++ b/src/components/vue-form-builder.vue
@@ -65,7 +65,7 @@
           <i class="far fa-trash-alt"/>
         </b-button>
 
-        <b-button size="sm" variant="secondary" class="ml-1" v-b-modal.addPageModal>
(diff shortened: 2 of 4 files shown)


nolanpro:
@chipit24 I updated with your suggestions and lint passes. I still don't understand why i'm getting `Error: Validator 'unique-page-name' is not defined!` and the validator doesn't work for me. This happens when I edit a page name (in processmaker, npm linked) but I suspect it has to do with externals. I know you said you didn't see the problem but could you check again? The standalone app works fine.

chipit24:
@nolanpro I tested it within processmaker and I'm getting the error you described. It does seem like a single validator instance will be used, so I set it up as an external / peer dependency, but that didn't fix the problem.

I also verified that the validation rules are registered before `new Validator` is called.

nolanpro:
@chipit24 ok figured it out, see https://github.com/ProcessMaker/vue-form-elements/pull/64

chipit24:
@nolanpro I pulled in both this branch and `vue-form-elements/bugfix/63` but I'm still seeing the same problem. Did you mean to set up `validatorjs` as an external on this branch as well?

nolanpro:
@chipit24 oops forgot to push the commit to this branch, just did, try it again

chipit24:
@nolanpro This PR looks good and can be merged after the conflicts are resolved.

PR #347, opened 2019-08-13
Computed Property Modal

### LINKED ISSUE #345: Computed Property Modal
The save/cancel buttons in this modal don’t follow our standard UI patterns. They should be flipped and only say Save and Cancel.

Also the “Property Saved” alert persists even when I go to add another Computed Property, as seen above. This should show in the main window, not in the modal.

Also, the validation errors and red outlines appear in this modal as soon as it’s opened, contrary to all of our other modals.

In conclusion, this modal is quite something.

Current Modal: 

image

Correct Modal Pattern: 

image

marcoAntonioNina:
Resolves #345 

- Fix style computed property modal

require ProcessMaker/vue-form-elements#64

### DIFF
--- a/src/components/computed-properties.vue
+++ b/src/components/computed-properties.vue
@@ -86,6 +86,7 @@
         :validation="rulePropName"
       />
       <form-text-area
+        ref="propDescription"
         v-model="add.name"
         :label="$t('Description')"
         :name="$t('Description')"
@@ -101,7 +102,9 @@
             <i class="fab fa-js-square"/>
           </a>
         </div>
+
         <form-text-area
+          ref="propFormula"
           v-show="!isJS"
           rows="5"
           v-model="add.formula"
@@ -198,9 +201,9 @@ export default {
     },
     'add.property': {
       handler() {
-        if (this.$refs.propName && this.$refs.propName.validator) {
+        /*if (this.$refs.propName && this.$refs.propName.validator) {
           this.$refs.propName.validator.messages.messages.not_in = this.$t('Property already exists.');
-        }
+        }*/
         let data = '';
         this.current.forEach(item => {
           if (item.property === this.add.property && item.id !== this.add.id) {
@@ -209,19 +212,17 @@ export default {
         });
         this.rulePropName = 'required' + (data ? '|' + data : '');
       },
-      deep: true,
     },
     'add.name': {
       handler() {
         this.ruleDescription = 'required';
       },
-      deep: true,
     },
     'add.formula': {
       handler() {
         this.ruleFormula = 'required';
+        this.editorInvalid = this.add.formula.trim() === '';
       },
-      deep: true,
     },
   },
   computed: {
@@ -277,18 +278,17 @@ export default {
     validateData() {
       this.ruleDescription = 'required';
       this.ruleFormula = 'required';
-      this.rulePropName = this.rulePropName ? this.rulePropName : 'required';
+      this.rulePropName = this.rulePropName || 'required';
       this.editorInvalid = this.add.formula.trim() === '';
-      let fields = this.$children[0].$children[0].$children;
 
-      let valid = false;
-      fields.forEach(item => {
-        if (item.name && item.validator) {
-          valid += item.validator.errorCount;
+      let valid = true;
+      for (let item in this.$refs) {
+        if (this.$refs[item].name && this.$refs[item].validator && this.$refs[item].validator.errorCount !== 0) {
+          valid = false;
         }
-      });
+      }
 
-      if (valid === 0) {
+      if (valid && !this.editorInvalid) {
         this.saveProperty();
       }
     },
--- a/src/components/computed-properties.vue
+++ b/src/components/computed-properties.vue
@@ -141,6 +141,8 @@ import {
 } from '@processmaker/vue-form-elements';
 import MonacoEditor from 'vue-monaco';
 
+let Validator = require('validatorjs');
+
 const globalObject = typeof window === 'undefined'
   ? global
   : window;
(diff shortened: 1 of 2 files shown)


src/components/computed-properties.vue
```diff
@@ -78,25 +73,28 @@
...
     <template v-else>
       <form-input
         v-model="add.property"
         :label="$t('Property Name')"
-        name="property name"
-        validation="required"
+        :name="$t('Property Name')"
+        :validation="rulesValidation"
       />
+      <div v-if="existsProperty" class="invalid-feedback d-block">
```
chipit24: I think it would be cleaner to add this as a custom validator rule:
https://github.com/skaterdav85/validatorjs#register-custom-validation-rules

src/components/computed-properties.vue
```diff
@@ -238,15 +259,18 @@ export default {
...
         if (item.property === this.add.property && item.id !== this.add.id) {
-          validation = false;
-          this.showAlert('Property already exists', 'danger');
+          this.existsProperty = true;
         }
       });
-      if (validation) {
+    },
+    validateData() {
+      this.rulesValidation = 'required';
```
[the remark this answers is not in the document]
marcoAntonioNina: I agree with you

src/components/computed-properties.vue
```diff
@@ -130,6 +137,10 @@ import {
 } from '@processmaker/vue-form-elements';
 import MonacoEditor from 'vue-monaco';
 
+let globalObject = typeof window === 'undefined'
```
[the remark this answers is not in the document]
marcoAntonioNina: Fixed.

src/components/computed-properties.vue
```diff
@@ -289,34 +313,36 @@ export default {
...
+      this.showAlert(this.$t('Property deleted'), 'success');
       this.displayTableList();
     },
     showAlert(message, variant) {
-      this.alertVariant = variant || 'success';
-      this.message = message || '';
-      this.showDismissibleAlert = true;
+      if (globalObject.ProcessMaker && globalObject.ProcessMaker.alert) {
+        globalObject.ProcessMaker.alert(message, variant || 'success');
+      }
```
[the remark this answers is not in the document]
marcoAntonioNina: Fixed

src/components/computed-properties.vue
```diff
@@ -259,18 +274,21 @@ export default {
...
     },
-    validateDuplicate() {
-      this.existsProperty = false;
-      this.current.forEach(item => {
-        if (item.property === this.add.property && item.id !== this.add.id) {
-          this.existsProperty = true;
+    validateData() {
+      this.ruleDescription = 'required';
+      this.ruleFormula = 'required';
+      this.rulePropName = this.rulePropName ? this.rulePropName : 'required';
```
chipit24: A ternary here isn't necessary; I think you can make it more readable by using a logical OR operator:
```js
this.rulePropName = this.rulePropName || 'required';
```

src/components/computed-properties.vue
```diff
@@ -259,18 +274,21 @@ export default {
...
+      this.editorInvalid = this.add.formula.trim() === '';
+      let fields = this.$children[0].$children[0].$children;
+
+      let valid = false;
```
chipit24: Since you end up treating this variable as a number, I think it would make more sense to initialize it to `0` instead of `false`.

src/components/computed-properties.vue
```diff
@@ -259,18 +274,21 @@ export default {
...
-      this.existsProperty = false;
-      this.current.forEach(item => {
-        if (item.property === this.add.property && item.id !== this.add.id) {
-          this.existsProperty = true;
+    validateData() {
+      this.ruleDescription = 'required';
+      this.ruleFormula = 'required';
+      this.rulePropName = this.rulePropName ? this.rulePropName : 'required';
+      this.editorInvalid = this.add.formula.trim() === '';
+      let fields = this.$children[0].$children[0].$children;
```
chipit24: This is a little hard to read (what exactly are these deep nested children?) and fragile. If this component's template is refactored, this code could break. I think you could use `$refs` here: https://vuejs.org/v2/api/#vm-refs. Adding the same ref string to multiple components will create an array for you.

Also, you can use `const` here instead of `let` since fields isn't re-assigned.

src/components/computed-properties.vue
```diff
@@ -195,7 +198,28 @@ export default {
...
+        }
+        let data = '';
+        this.current.forEach(item => {
+          if (item.property === this.add.property && item.id !== this.add.id) {
+            data += data ? ',' + item.property : 'not_in:' + item.property;
+          }
+        });
+        this.rulePropName = 'required' + (data ? '|' + data : '');
+      },
+      deep: true,
```
chipit24: I don't think you need to specify `deep: true` for any of the `add` property watchers; all of these properties are primitive values (not objects), so there's no actual _deep_ value to watch.

src/components/computed-properties.vue
```diff
@@ -195,7 +198,28 @@ export default {
     },
     'add.property': {
       handler() {
-        this.validateDuplicate();
+        if (this.$refs.propName && this.$refs.propName.validator) {
```
chipit24: Using a ref to reach into a component's internal state is a code smell.

It's hard to read: I have to now look into the `form-input` component to get a complete picture of how this code is suppose to work.

It's fragile: If the component the ref is referring to is refactored, then this code might break; it introduces a tight coupling.

There has to be a better way to handle this without introducing this kind of tech debt. What do you think?

src/components/computed-properties.vue
marcoAntonioNina: fixed

src/components/computed-properties.vue
marcoAntonioNina: fixed

src/components/computed-properties.vue
marcoAntonioNina: changed

src/components/computed-properties.vue
marcoAntonioNina: Changed

src/components/computed-properties.vue
marcoAntonioNina: yes, I changed to add validation rule with pull request in  project vue-form-elements

required ProcessMaker/vue-form-elements#64

PR #348, opened 2019-08-13
Hide toolbar in typeDisplay

### LINKED ISSUE #344: Display Screen Pages
Display type screens DO NOT have multiple pages. Pages should be hidden.

tuantruong625:
**Type Form**
Screen Shot 2019-08-13 at 2 08 34 PM

**Type Display**
Screen Shot 2019-08-13 at 2 07 56 PM

PR in processmaker repo is required to be merge.
https://github.com/ProcessMaker/processmaker/pull/2275

Fixes #344

### DIFF
--- a/src/components/vue-form-builder.vue
+++ b/src/components/vue-form-builder.vue
@@ -42,36 +42,38 @@
     <!-- Renderer -->
     <b-col class="overflow-auto mh-100 ml-4 mr-4 p-0 d-flex flex-column position-relative pt-2">
       <b-input-group size="sm" class="bg-white mt-3">
-        <b-form-select v-model="currentPage" class="form-control">
+        <b-form-select v-if="!showToolbar" v-model="currentPage" class="form-control">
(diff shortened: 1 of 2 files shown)


src/components/vue-form-builder.vue
```diff
@@ -358,6 +361,12 @@ export default {
     isCurrentPageEmpty() {
       return this.config[this.currentPage].items.length === 0;
     },
+    showToolbar() {
+      if (!this.screenType) {
+        return;
```
chipit24: This reads backwards to me. If `showToolbar` is `false`, the toolbar is shown. I'd assume that `showToolbar === true` would mean the toolbar should be shown.

When you don't specify a return value for a function, it implicitly returns `undefined`, however, I think in this case it should always return a boolean. So the above line could be `return false;`.

I'd change the whole thing to something like:

```
return this.screenType === formTypes.display;
```

And ensure that the `screenType` prop gets a default value of `formTypes.display` if it's not set.

PR #369, opened 2019-08-28
FIx custom css scoping

### LINKED ISSUE #350: Add CSS Scope to Forms
Add a scope to the Custom CSS to be applied to the entire form, not just an element. We never want a custom css rule to bleed out of the container. Look at scoped css with vue. This scope would be available in the task detail, web entry forms and the screen builder preview.

### LINKED ISSUE #353: Custom CSS Should Not Show in Editor View
The CSS scoping is definitely wrong as is the css showing in the edit view

image

nolanpro:
Fixes #350 
Fixes #353


chipit24:
Looks like there was only 1 eslint error that I had to fix manually.

velkymx:
.form-group:nth-of-type(1){
  background-color: chartreuse
}

chipit24:
> .form-group:nth-of-type(1){
> background-color: chartreuse
> }

I tested it with this CSS specifically as well, and the original issue appears to be fixed.

chipit24:
Looks like I missed something with this. When testing these changes in modeler, it looks like the following change:
https://github.com/ProcessMaker/screen-builder/pull/369/files#diff-e9ee7bdaa20db1e8d5c6926389e68f3aR3

Has caused extra padding to appear around the modelers inspector control's:
Screen Shot 2019-08-29 at 1 58 53 PM

nolanpro:
@chipit24 will fix in separate issue, thanks

PR #370, opened 2019-08-28
Move custom css selector attr back to the parent element

### LINKED ISSUE #355: Move Selector
In this example (which is incorrect) the selector "fullname" is applied to the input field. 

image

The select needs to be applied to the div container, in this case it's the div with the class form-group.

Example

```<div class="form-group" selector="fullname">```

nolanpro:
Fixes #355


nolanpro:
@chipit24 fix it in this branch

PR #378, opened 2019-08-29
Added null checking to avoid error when adding a new page.

### LINKED ISSUE #375: When saving a new page or switching between pages, it displays a console error
**Steps to reproduce:**
1. On a form screen, add a new page
2. In the page dropdown, switch between pages

Example: https://www.dropbox.com/s/p54e1apst28ocpb/open-of-undefined.mov?dl=0

**Expected:**
There should be no errors in the console log on page save. There should be no additional console errors every time you switch the page in the dropdown.

**Actual:**
It prints these errors in the console log: 

> Error in callback for watcher "currentPage": "TypeError: Cannot set property 'open' of undefined"
Screen Shot 2019-08-29 at 11 57 42 AM

Seems like it is related to this line added in https://github.com/ProcessMaker/screen-builder/pull/362: 
https://github.com/ProcessMaker/screen-builder/pull/362/commits/c31c027633d95cd5457c180c7a54b96cecd6d0ee#diff-f631b3a2a589629965342504efc692b7R425

JCook21:
Fixes #375. This PR adds null checking for the result of an Array.prototype.find() operation, which is not guaranteed to return a result.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -900,7 +900,7 @@
     "@intervolga/optimize-cssnano-plugin": {
       "version": "1.0.6",
       "resolved": "https://registry.npmjs.org/@intervolga/optimize-cssnano-plugin/-/optimize-cssnano-plugin-1.0.6.tgz",
-      "integrity": "sha1-vnx4RhKLiPapsdEmGgrQbrXA/fg=",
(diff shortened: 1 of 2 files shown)


JCook21:
@velkymx @lisa-fehr @caleeli package-lock.json was updated when I ran `npm install`. I left it in this PR but I can revert it if you prefer.

PR #385, opened 2019-09-02
Read Only Controls

### LINKED ISSUE #384: Multi-Column Read Only Controls
Story: Sometime I want to be able to display the form elements without allowing the user to make changes to them.

This functionality is already available in the input and textarea controls, but needs to be added to:

- [x] Selects
- [x] Radio Groups (so you can't change the selected radio
- [x] Date Picker
- [x] Checkbox

Add a checkbox to make a form element read only (see UI #249)

https://getbootstrap.com/docs/4.1/components/forms/#readonly

marcoAntonioNina:
Resolves #384 

Bootstrap does not introduce special considerations for disabling Select, radio groups, checkbox.

- Add disabled property in Select, Radio Groups, Date Picker, checkbox.

### DIFF
--- a/src/components/vue-form-renderer.vue
+++ b/src/components/vue-form-renderer.vue
@@ -31,7 +31,7 @@
             :name="element.config.name !== undefined ? element.config.name : null"
             v-bind="element.config"
             :is="element.component"
-            :disabled="element.config.interactive"
+            :disabled="element.config.interactive ? element.config.interactive : element.config.disabled"
           />
         </div>
       </div>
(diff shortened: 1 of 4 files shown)


src/components/vue-form-renderer.vue
```diff
@@ -31,7 +31,7 @@
             :name="element.config.name !== undefined ? element.config.name : null"
             v-bind="element.config"
             :is="element.component"
-            :disabled="element.config.interactive"
+            :disabled="element.config.interactive ? element.config.interactive : element.config.disabled"
```
chipit24: You can use an _or_ statement here:
```
:disabled="element.config.interactive || element.config.disabled"
```

marcoAntonioNina: fixed

PR #404, opened 2019-09-26
Record List Improvements

### LINKED ISSUE #399: Records List Improvement
- [x] Create an specific component for Column Setup
- [x] Make Accordion Header "Columns"
- [x] Form Header - Add/Edit Column 
- [x] Value/Content change to Column Header and Value
- [x] Column value needs to support mustache for concatenated columns
- [x] BUG: On list view , the edit / delete buttons touch (and the delete button should be red and the bin icon)
- [x] Remove JSON Options from the record list

image

caleeli:
Resolves #399 

image

### DIFF
--- a/src/components/accordions.js
+++ b/src/components/accordions.js
@@ -33,16 +33,11 @@ export default [
     open: false,
   },
   {
-    name: 'Data Source',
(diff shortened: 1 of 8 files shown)


caleeli:
@nolanpro I fixed the value rendered when the field is empty

PR #413, opened 2019-10-07
Set default values for variables

### LINKED ISSUE #351: Default Values
All variables should have a default value. This will guarantee that the variable is always available, even if it is set to null. 

Special case: Date / Time picker should default to NOW

If you submit a form the data object should contain the variable with no value. 

This is wrong:

image

this is correct

image

sanjacornelius:
closes #351 

**Issue:**
Data preview does not display variables with the value of null.

**Fix:**
This PR runs the 'setDefaultValues' method to display the variable's default value in the data preview.

Screen Shot 2019-10-07 at 2 40 30 PM

### DIFF
--- a/src/components/vue-form-renderer.vue
+++ b/src/components/vue-form-renderer.vue
@@ -301,7 +301,7 @@ export default {
       }
 
       if (item.component === 'FormDatePicker') {
-        let date = new Date();
+        const date = new Date();
         defaultValue = date.toISOString();
       }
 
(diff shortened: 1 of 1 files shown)


src/components/vue-form-renderer.vue
```diff
@@ -299,6 +300,11 @@ export default {
         defaultValue = [];
       }
 
+      if (item.component === 'FormDatePicker') {
+        let date = new Date();
```
chipit24: Since you're not re-assigning `date`, you can use `const` instead of `let`.

sanjacornelius: New commits include removing the `defaultValues` data key and updating `let` to `const` for the `date` variable. 😃

PR #418, opened 2019-10-14
Required Field Validation Issues

### LINKED ISSUE #414: Required Field Validation Issues
A Form with more than one page, with required fields in the first page(s) and a Submit button in the last page, will allow the Submit even if the required fields are not filled out. Validation seems to only be enforced in the actual page that holds the Submit button, not the entire form.

To recreate:
1. create a form screen
2. add line input. Add 'required' as Validation
3. add a new page, Page 2
4. add a Submit button to Page 2
5. back in Page 1, add a Navigation button into Page 2.
6. Test in preview. Do not fill the required field in Page 1, rather press the navigation button and submit from Page 2. Notice the form gets submitted.

marcoAntonioNina:
Resolves #414 

- Fix of validations on all pages of the screen.


src/components/vue-form-renderer.vue
```diff
@@ -73,6 +73,8 @@ Vue.component('custom-css', {
 
 Vue.use(VueDeepSet);
 
+let Validator = require('validatorjs');
```
chipit24: Nitpick: Since `Validator` isn't re-assigned, you can use `const` instead of `let`.

src/components/vue-form-renderer.vue
```diff
@@ -205,42 +207,41 @@ export default {
...
+          rules[`${item.config.name}`] = item.config.validation;
         }
       });
+
     },
     isValid() {
       this.errors = [];
-      this.valid = true;
+      let data = {};
+      let rules = {};
```
chipit24: Nitpick: You can use `const` for these two declarations as well.

src/components/vue-form-renderer.vue
```diff
@@ -205,42 +207,41 @@ export default {
...
+      this.getRules(this.config, data, rules);
```
chipit24: I'm not a fan of having function calls mutate the arguments it's applied to. It makes it hard to track the changes to these arguments, and hence makes it difficult to debug if there are issues.

I think having it return the new data and rules, like in the example below, will make this function easier to reason about, debug, and test:
```js
const { data, rules } = this.getRules(this.config);
```

And since it seems to return data _and_ rules, maybe it can be called `getDataAndRules`.

marcoAntonioNina: fixed

src/components/vue-form-renderer.vue
```diff
@@ -205,42 +207,45 @@ export default {
         this.$emit('submit', this.transientData);
       }
     },
-    validateElements(elements) {
-      elements.forEach(element => {
-        if (element.validator && element.validator.errorCount !== 0) {
-          this.valid = false;
-          this.errors.push(element.validator.errors.errors);
+    getDataAndRules(items) {
+      let data = {}, rules = {};
```
chipit24: You've taken the arguments and moved them into the method, but you should also be thinking about what the method is actually doing with them. You've replaced one problem with another: Instead of recursively passing down props by reference, you're now holding onto three different versions of the props that you spread into each other.

src/components/vue-form-renderer.vue
```diff
@@ -205,42 +207,45 @@ export default {
...
+
+        if (item.items) {
+          const [data2, rules2] = this.getDataAndRules(item.items);
+          data = {...data, ...data2};
+          rules = {...rules, ...rules2};
+        }
+
+        if (item.config && item.config.name && item.config.validation) {
+          data[`${item.config.name}`] = this.data[item.config.name];
+          rules[`${item.config.name}`] = item.config.validation;
```
chipit24: It looks like this loop is doing three different things. Two of them look the same (the `Array.isArray(item)` and `item.items` `if` blocks seem to do the same thing); so there's possibly some duplication that can be eliminated.

See if you can split it up into three different `forEach` loops to improve readability. Once you have it broken down, you may also find there's a nice way to refactor it back into a single method.

Think about readability and testability. This  method may also be moved to a validator factory, as I suggest in my comment below.

src/components/vue-form-renderer.vue
```diff
@@ -205,42 +207,45 @@ export default {
...
+      return [data, rules];
     },
     isValid() {
       this.errors = [];
-      this.valid = true;
+      const [ data, rules ] = this.getDataAndRules(this.config);
 
-      if (this.$refs && this.$refs.elements) {
-        this.validateElements(this.$refs.elements);
+      this.dataTypeValidator = new Validator( data, rules, null);
```
chipit24: Looking at the docs, it seems the 3rd argument here is an optional `customErrorMessages` object; you can leave this argument empty instead of passing `null`.

src/components/vue-form-renderer.vue
```diff
@@ -205,42 +207,45 @@ export default {
...
+          rules[`${item.config.name}`] = item.config.validation;
         }
       });
+
+      return [data, rules];
     },
     isValid() {
       this.errors = [];
-      this.valid = true;
+      const [ data, rules ] = this.getDataAndRules(this.config);
```
chipit24: Since these values are only being passed to `Validator`, an even better name might be `getDataAndRulesForValidator`.

However, this method has two dependencies, `this.config`, and `this.data`. It would be ideal to put this into a Validator factory, and have that manage this logic. It would also make it much easier to test.

The end result would look something like this, without a direct call to `getDataAndRules`:
```js
this.dataTypeValidator = ValidatorFactory(this.config, this.data);
```

chipit24:
I didn't re-test the code but it seems the functionality is unchanged.

> When I test this, it seems the issue still exists. I tested in both standalone and `npm link`ed into processmaker. Can you provide the steps you took to test this, and what I should expect?

@marcoAntonioNina Did you have a change to read over this comment in my original review?

marcoAntonioNina:
@chipit24 I added the factory and thanks for the recommendations

tests/unit/ValidationFactory.spec.js
```diff
@@ -0,0 +1,47 @@
...
...y'},{'value':'text-success','content':'success'},{'value':'text-danger','content':'danger'},{'value':'text-warning','content':'warning'},{'value':'text-info','content':'info'},{'value':'text-light','content':'light'},{'value':'text-dark','content':'dark'}]}},{'type':'ColorSelect','field':'bgcolor','config':{'label':'Background Color','helper':'Set the element\'s background color','options':[{'value':'alert alert-primary','content':'primary'},{'value':'alert alert-secondary','content':'secondary'},{'value':'alert alert-success','content':'success'},{'value':'alert alert-danger','content':'danger'},{'value':'alert alert-warning','content':'warning'},{'value':'alert alert-info','content':'info'},{'value':'alert alert-light','content':'light'},{'value':'alert alert-dark','content':'dark'}]}},{'type':'FormInput','field':'conditionalHide','config':{'label':'Visibility Rule','helper':'This control is hidden until this expression is true'}},{'type':'FormInput','field':'customCssSelector','config':{'label':'CSS Selector Name','helper':'Use this in your custom css rules','validation':'regex: [-?[_a-zA-Z]+[_-a-zA-Z0-9]*]'}}],'editor-control':'MultiColumn','editor-component':'MultiColumn'}]}];
```
chipit24: I think this _really_ long line would be better in a separate file, like `testConfig.json`.

marcoAntonioNina: I added the file with screen config

PR #422, opened 2019-10-16
Prevent default value being set on accordion

chipit24:
Fixes https://github.com/ProcessMaker/modeler/issues/743.

Also refactored and extracted code to make maintenance simpler as well as applying a linting format to the files affected.


src/itemProcessingUtils.js
```diff
@@ -0,0 +1,69 @@
+function removeInvalidOptions(option) {
+  return Object.keys(option).includes('value', 'contemnt') &&
```
JCook21: This is an interesting typo. It was in the code before so I didn't alter it. I'm pretty sure this code can't be working with the typo, should I change it now?

src/components/vue-form-renderer.vue
```diff
@@ -44,24 +44,25 @@
 import Vue from 'vue';
 import * as VueDeepSet from 'vue-deepset';
 import debounce from 'lodash/debounce';
-import { HasColorProperty, shouldElementBeVisible, getValidPath } from '@/mixins';
+import {getValidPath, HasColorProperty, shouldElementBeVisible} from '@/mixins';
```
lisa-fehr: Should we return the spacing between the braces in this file?

src/itemProcessingUtils.js
chipit24: You can change it. `includes` only accepts one argument, so `'contemnt'` can be removed. I'm fairly confident that what this code should be doing is checking that `option` is an object with `value` and `content` keys, and that the `content` key is not `null` or `undefined`.

lisa-fehr: There are at least 3 files with this spelling mistake. I wonder.

src/components/vue-form-renderer.vue
JCook21: Yeah, it could be a setting that's missing in eslint or my config for this project. Let me take a look.

src/itemProcessingUtils.js
chipit24: The other two files are in a different repo/project though (in `vue-form-elements`). And it appears the setup is the same; `includes` only accepts one argument so it's doing nothing in all cases.

JCook21: I could add a second includes call, checking for content (with correct spelling) ...

src/itemProcessingUtils.js
```diff
@@ -0,0 +1,69 @@
...
+
+export function getDefaultValueForItem(item) {
+  let defaultValue = null;
+
+  if (['FormInput', 'FormTextArea', 'FormText'].includes(item.component)) {
+    defaultValue = '';
+  }
+
+  if (['FormSelect', 'FormRadioButtonGroup'].includes(item.component) && item.config.options) {
+    const options = getOptionsFromDataSource(item.config.options, this.transientData);
```
lisa-fehr: Think I had an issue with `this.transientData`

src/components/vue-form-renderer.vue
```diff
@@ -244,57 +206,18 @@ export default {
...
-
-      let defaultValue = null;
-
-      if (['FormInput', 'FormTextArea', 'FormText'].includes(item.component)) {
-        defaultValue = '';
-      }
-
-      if (['FormSelect', 'FormRadioButtonGroup'].includes(item.component) && item.config.options) {
-        const options = getOptionsFromDataSource(item.config.options, this.transientData);
+      const shouldHaveDefaultValue = item => {
```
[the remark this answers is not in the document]
JCook21: Could be either. I wanted to pull this out of the component completely but the dependency on the model and `getValidPath()` methods stopped that. I was passing it as a callback, hence the function. I left it like this as it doesn't really 'belong' in the class and has no real links to it.

src/itemProcessingUtils.js
lisa-fehr: Looks like it relates to the radio button group

JCook21: @lisa-fehr in what way?

JCook21: @lisa-fehr can you post complete reproduction steps and I'll take a look into it?

lisa-fehr: @JCook21 in that this.transientData is from the old file.

It says
```
TypeError: Cannot read property 'transientData' of undefined
    at getDefaultValueForItem (itemProcessingUtils.js?504e:51)
```

When I add a select or radio group to screen builder after giving them variable names and clicking preview, I get that error above.

src/itemProcessingUtils.js
chipit24: @JCook21 Because of the `option.content != null` check, adding that 2nd includes would be redundant. If there's no `content` key, then `option.content` would be `undefined` and so `option.content != null` would return `false`.

So I think it's enough to only remove `'contemnt'`.

src/itemProcessingUtils.js
JCook21: Ahh, good catch. Fixed. :)

lisa-fehr: For radio group, you must select the Data Type to see the error.

PR #427, opened 2019-10-21
Add cypress

chipit24:


chipit24:
> * Could we go the extra mile and add setup for code coverage collection too?

I tried this and went down a rabbit hole, fixing one problem and then having another one crop up. I think code coverage is complicated enough that it can be added in a separate PR without holding these changes up.

> * How much value is there in having so much commented out code in this PR? Should we just remove all, or the majority, of it?

I removed the comments. It was mainly code snippets and help text; but that's why there's docs.cypress.io.

> * Can you update the README with information about Cypress?

I'll update the README!

PR #447, opened 2019-11-25
Remove Datatype dropdown 

### LINKED ISSUE #440: Checkbox Control
This can only be a boolean. Remove the data type drop down from the inspector

tuantruong625:
The purpose of this pull request is to remove the Data Type dropdown from the Checkbox inspector.

**Acceptance Criteria** 
The checkbox inspector panel does not contain DataType Dropdown.

Screen Shot 2019-11-25 at 2 15 54 PM

Fixes #440

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15629,9 +15629,9 @@
       "dev": true
     },
     "vue-loader": {
-      "version": "15.7.1",
-      "resolved": "https://registry.npmjs.org/vue-loader/-/vue-loader-15.7.1.tgz",
-      "integrity": "sha512-fwIKtA23Pl/rqfYP5TSGK7gkEuLhoTvRYW+TU7ER3q9GpNLt/PjG5NLv3XHRDiTg7OPM1JcckBgds+VnAc+HbA==",
+      "version": "15.7.2",
+      "resolved": "https://registry.npmjs.org/vue-loader/-/vue-loader-15.7.2.tgz",
+      "integrity": "sha512-H/P9xt/nkocyu4hZKg5TzPqyCT1oKOaCSk9zs0JCbJuy0Q8KtR0bjJpnT/5R5x/Ckd1GFkkLQnQ1C4x6xXeLZg==",
       "requires": {
         "@vue/component-compiler-utils": "^3.0.0",
         "hash-sum": "^1.0.2",
--- a/package-lock.json
+++ b/package-lock.json
@@ -15629,9 +15629,9 @@
       "dev": true
     },
     "vue-loader": {
-      "version": "15.7.2",
-      "resolved": "https://registry.npmjs.org/vue-loader/-/vue-loader-15.7.2.tgz",
-      "integrity": "sha512-H/P9xt/nkocyu4hZKg5TzPqyCT1oKOaCSk9zs0JCbJuy0Q8KtR0bjJpnT/5R5x/Ckd1GFkkLQnQ1C4x6xXeLZg==",
+      "version": "15.7.1",
+      "resolved": "https://registry.npmjs.org/vue-loader/-/vue-loader-15.7.1.tgz",
+      "integrity": "sha512-fwIKtA23Pl/rqfYP5TSGK7gkEuLhoTvRYW+TU7ER3q9GpNLt/PjG5NLv3XHRDiTg7OPM1JcckBgds+VnAc+HbA==",
       "requires": {
         "@vue/component-compiler-utils": "^3.0.0",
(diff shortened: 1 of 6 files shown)


package-lock.json
```diff
@@ -15629,9 +15629,9 @@
       "dev": true
     },
     "vue-loader": {
-      "version": "15.7.1",
-      "resolved": "https://registry.npmjs.org/vue-loader/-/vue-loader-15.7.1.tgz",
-      "integrity": "sha512-fwIKtA23Pl/rqfYP5TSGK7gkEuLhoTvRYW+TU7ER3q9GpNLt/PjG5NLv3XHRDiTg7OPM1JcckBgds+VnAc+HbA==",
```
chipit24: Do you think this change to the lockfile can be reverted? The lockfile usually should only change when dependencies are changed (added, removed, updated).

chipit24:
Something to note with this change: If a user has a form saved with the previous checkbox with the datatype attribute not set, it will show up as an error with no way to fix it unless the checkbox is deleted and re-added.
Screen Shot 2019-11-26 at 10 12 12 AM
I don't think this will be an issue, but just something to note.

chipit24:
Take a look at my comment in your other PR regarding the data format mixin: https://github.com/ProcessMaker/screen-builder/pull/448#issuecomment-558686014. Even though the default validation is to convert the value to a string if data format is missing, it didn't seem to break the checkbox (it still set the value as a boolean instead of a string).

JCook21:
Same comment as in #448: this needs unit and/or integration tests.

tests/unit/ConditionalHide.spec.js
```diff
@@ -60,43 +60,43 @@ describe('Hide or show controls', () => {
   it('Hide control with evaluation success', () => {
 
     //The fields that not contain a conditional are displayed by default
-    expect(wrapper.vm.showElement.field1).toBe(true);
-    expect(wrapper.vm.showElement.field4).toBe(true);
+    expect(wrapper.contains('#field1')).toBe(true);
```
tuantruong625: There was also another option of using `expect(wrapper.html()).toContain('<div id="field1"></div>....)` but I thought doing this way would keep the test simipler.

src/components/inspector/options-list.vue
```diff
@@ -332,27 +331,27 @@ export default {
...
-      }	
-      catch (err) {	
-        this.jsonError = err.message;	
-        return;	
+    jsonDataChange() {
+      let jsonList = [];
+      try {
+        jsonList = JSON.parse(this.jsonData);
+        if (jsonList.constructor !== Array && jsonList.constructor !== Object) {
+          throw Error('String does not represent a valid JSON');
```
JCook21: The try catch block is superfluous here. Inside of this if block why not just assign the error message to `this.jsonError` and return?

src/components/inspector/options-list.vue
```diff
@@ -332,27 +331,27 @@ export default {
...
+      catch (err) {
+        this.jsonError = err.message;
+        return;
+      }
+      this.optionsList = [];
+      const that = this;
+      jsonList.forEach (item => {
+        that.optionsList.push({
+          [that.keyField] : item[that.keyField],
+          [that.valueField] : item[that.valueField],
```
JCook21: Isn't the `that=this` business unneeded in ES6? Also couldn't this be done using a map which would simplify it, something like:
```suggestion
     this.optionsList = jsonList.map(item => {
         return {
             [this.keyField] : item[this.keyField],
             [this.valueField] : item[this.valueField]
     });
```

src/components/inspector/options-list.vue
tuantruong625: I just removed the require statement that was throwing an error which caused the test to fail and the linter formatted the code. I can do the refactor as part of this PR?

src/components/inspector/options-list.vue
```diff
@@ -332,27 +331,27 @@ export default {
...
-          throw Error('String does not represent a valid JSON');	
-        }	
-      }	
-      catch (err) {	
-        this.jsonError = err.message;	
-        return;	
+    jsonDataChange() {
+      let jsonList = [];
+      try {
+        jsonList = JSON.parse(this.jsonData);
```
JCook21: JSON.parse throws a SyntaxError if the string is not valid JSON. Would this be enough error checking?

tuantruong625:
jsonDataChange() method will require some refactoring as discussed with @JCook21. Will refactor after once this PR is merged.

PR #448, opened 2019-11-25
Textarea Element: Remove Dataype dropdown

### LINKED ISSUE #438: Textarea Data Type: Text
The text area component should except text type of data. Remove the option to set the data type for this control.

tuantruong625:
The purpose of this pull request is to remove the datatype down from the inspector.

**Acceptance Criteria.**
The text area inspector should not contain the datatype select.

Screen Shot 2019-11-25 at 2 31 28 PM

Fixes #438

### DIFF
--- a/src/form-builder-controls.js
+++ b/src/form-builder-controls.js
@@ -133,7 +133,6 @@ export default [
         labelProperty,
         placeholderProperty,
         keyNameProperty,
-        DataTypeProperty,
         helperTextProperty,
         {
           type: 'FormCheckbox',
--- a/tests/unit/VueFormBuilder.spec.js
+++ b/tests/unit/VueFormBuilder.spec.js
@@ -23,6 +23,10 @@ function isFormCheckbox(config) {
   return config.control.component === 'FormCheckbox';
 }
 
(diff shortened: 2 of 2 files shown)


chipit24:
The comment I left at https://github.com/ProcessMaker/screen-builder/pull/447#issuecomment-558675997 applies here as well.

This change may cause problems later on; even though the data type dropdown has been removed, the `FormTextArea` component still contains the `DataFormatMixin` mixin, which will format the value of the textarea based on its `dataFormat` property. 

Since you removed this dropdown from the component, I think it makes sense to remove the `DataFormatMixin` mixin as well. You may want to do this for the `FormCheckbox` component as well.

JCook21:
We cannot make changes to the behaviour of this code without adding tests. Before making changes we need to add tests that guarantee changes do not break the behaviour unintentionally. Can you please add some unit and/or integration tests for this code?

JCook21:
The changes that @chipit24 mentioned are in The Vue-form-elements repo, not this one, so I'm going to go ahead and merge this PR. I'll look into the changes suggested in vue form elements.

PR #457, opened 2019-12-05
Watcher Script Config Not Saving

### LINKED ISSUE #456: Watcher Script Config Not Saving
https://github.com/ProcessMaker/processmaker/issues/2643

marcoAntonioNina:
Closes #456 

## Changes
- validation of which panel is active.
- validation json

## Test
- in processmaker create screen, config watcher then save and reload information.

## Screenshot
image

### DIFF
--- a/src/components/watchers-form.vue
+++ b/src/components/watchers-form.vue
@@ -20,6 +20,9 @@
       :validation="ruleWatcherVariable"
       @open="loadVariables"
     />
+    <div v-if="!config.watching" class="invalid-feedback d-block">
+      <div>{{ $t('The Variable to Watch field is required') }}</div>
+    </div>
 
     <form-multi-select
       :name="$t('Script Source')"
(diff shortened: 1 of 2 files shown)


src/components/watchers-form.vue
```diff
@@ -329,7 +356,20 @@ export default {
...
+        !this.jsonIsValid('script_configuration')
+      ) {
+        valid = false;
+      }
+
+      if (!valid && globalObject.ProcessMaker && globalObject.ProcessMaker.alert) {
+        globalObject.ProcessMaker.alert(this.$t('An error occurred. Check the form for errors in red text.'), 'danger');
+      }
+
+      if (valid) {
```
chipit24: I think this method can be refactored a bit. The `valid` variable it set in two places, and I think the logic for that could be broken out into a separate method, making it easier to use an early return. I'm thinking of something like this:

```js
if (this.isFormValid()) {
  // show alert
  return;
}

this.save();
```

A name like `validateDataAndSave` instead of `validateData` may also better communicate what this function is doing.

marcoAntonioNina: I like the idea

marcoAntonioNina: @chipit24 the changes have already been made.

PR #462, opened 2019-12-09
Watcher continually updating

### LINKED ISSUE #461: Watcher continually updating
https://github.com/ProcessMaker/processmaker/issues/2645

marcoAntonioNina:
closes #461 

## Changes
Add validation in the Check Watcher function, the array is ​​the same

### DIFF
--- a/src/mixins/formWatchers.js
+++ b/src/mixins/formWatchers.js
@@ -1,5 +1,5 @@
 import Mustache from 'mustache';
-import _ from 'lodash';
+import isEqual from 'lodash/isEqual';
 
 const globalObject = typeof window === 'undefined'
   ? global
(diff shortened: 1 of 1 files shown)


src/mixins/formWatchers.js
```diff
@@ -30,8 +30,7 @@ export default {
      * @param {object} data
      */
     checkWatcher(watcher, data) {
-      const trigger = _.get(this.watching, watcher.watching) != _.get(data, watcher.watching);
-      if (trigger) {
+      if (!_.isEqual(_.get(this.watching, watcher.watching), _.get(data, watcher.watching))) {
```
chipit24: Instead of importing a reference to the entire lodash library, I think it would be cleaner to import only the functions you need:
```js
import isEqual from 'lodash/isEqual';
```

This is what we do in modeler and it helps reduce bundle size. See the 2nd paragraph of https://lodash.com/per-method-packages.

Using `get` seems unnecessary here. I think something like this should work:
```js
if (isEqual(this.watching[watcher.watching], data[watcher.watching])) { ...
```

chipit24: I didn't look too much into how this code works, but I wonder if `isEqual` is necessary. Do you think this could be accomplished using vue's build-in `$watch` functionality? https://vuejs.org/v2/api/#vm-watch.

marcoAntonioNina: @chipit24 the change was made for the importation of lodash

PR #464, opened 2019-12-11
Watchers Running without Variable Interaction

### LINKED ISSUE #463: Watchers Running without Variable Interaction
https://github.com/ProcessMaker/processmaker/issues/2642

marcoAntonioNina:
closes #463 

## Changes
verify changes in the data for the watchers call

### DIFF
--- a/src/components/vue-form-renderer.vue
+++ b/src/components/vue-form-renderer.vue
@@ -148,7 +148,12 @@ export default {
 
         if (JSON.stringify(this.transientData) !== JSON.stringify(this.data)) {
           this.$emit('update', this.transientData);
-          this.watchDataChanges(this.transientData);
+
+          //Verify changes in the Data for the watchers call
+          let diff = this.dataChanges(this.transientData, this.data);
+          if (diff.length) {
+            this.watchDataChanges(this.transientData, diff);
+          }
         }
       },
       deep: true,
@@ -168,6 +173,23 @@ export default {
     }
   },
   methods: {
+    /**
+     * Returns the names of the fields that have changed.
+     *
+     * @param {object} newData
+     * @param {object} oldData
+     * @returns {[]}
+     */
+    dataChanges(newData, oldData) {
+      let changes = [];
+      for (const field in newData) {
+        if (oldData[field] !== undefined &&
+          JSON.stringify(newData[field]) !== JSON.stringify(oldData[field])) {
+          changes.push(field);
+        }
+      }
+      return changes;
+    },
     submit() {
       if (this.isValid()) {
         this.setDefaultValues();
--- a/src/mixins/formWatchers.js
+++ b/src/mixins/formWatchers.js
@@ -1,5 +1,6 @@
 import Mustache from 'mustache';
 import isEqual from 'lodash/isEqual';
+import get from 'lodash/get';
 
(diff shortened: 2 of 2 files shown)


src/mixins/formWatchers.js
```diff
@@ -21,10 +22,15 @@ export default {
     /**
      * Watch data changes
      * @param {object} data
+     * @param {array} changes
      */
-    watchDataChanges(data) {
+    watchDataChanges(data, changes) {
       if (this.watchers && this.watchers instanceof Array) {
```
chipit24: I think using `Array.isArray(this.watchers)` is more readable than an `instanceof` check. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray. So it should be possible to re-write this `if` block as:

```js
if (Array.isArray(this.watchers)) { ...
```

The initial check for `this.watchers` is unnecessary since a falsey value is not an array and would be caught by the `isArray` check.

src/mixins/formWatchers.js
```diff
@@ -21,10 +22,15 @@ export default {
...
-        this.watchers.forEach(watcher => this.checkWatcher(watcher, data));
+        this.watchers.forEach(watcher =>  {
+          if (changes.indexOf(watcher.watching) !== -1) {
```
chipit24: Use `changes.includes(watcher.watching)` instead of `indexOf` as it's more readable: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes#:~:targetText=The%20includes()%20method%20determines,true%20or%20false%20as%20appropriate.

src/components/vue-form-renderer.vue
```diff
@@ -148,7 +148,12 @@ export default {
 
         if (JSON.stringify(this.transientData) !== JSON.stringify(this.data)) {
           this.$emit('update', this.transientData);
-          this.watchDataChanges(this.transientData);
+
+          //Verify changes in the Data for the watchers call
+          let diff = this.dataChanges(this.transientData, this.data);
```
chipit24: `diff` could be instantiated with `const` instead of `let` since it doesn't seem to be re-assigned.

src/components/vue-form-renderer.vue
```diff
@@ -168,6 +173,23 @@ export default {
...
   },
   methods: {
+    /**
+     * Returns the names of the fields that have changed.
+     *
+     * @param {object} newData
+     * @param {object} oldData
+     * @returns {[]}
+     */
+    dataChanges(newData, oldData) {
```
chipit24: `dataChanges` returns an array of object keys for the properties that are **not** `undefined` and that are different between `newData` and `oldData`. The comment above takes up a large block and comments tend to become stale as the code changes, creating confusion.

In my opinion, comments are part of the code, and need to be maintained, but tend not to be, as the code changes. And an incorrect comment is more dangerous than no comment at all. And, if the need arises to examine this code, you will have to read through the code regardless, and not rely on a comment.

When skimming through the code, an accurate, descriptive name will provide more value than a comment, especially since the comment won't be pasted everywhere the function is called.

My suggestion is to remove the comment and rename the function to something like `getKeysForChangedObjectValues`.

PR #465, opened 2019-12-11
Data Object Missing from Select Lists

### LINKED ISSUE #455: Data Object Missing from Select Lists
When using a Select List in the screen builder there is no way to use an existing array. Instead you are presented with the Data Connector config.

Fix:
Keep the current config for data object, but rename it to be Data Connector.

Return the Data Object as defined as the use of an existing array generate via the request.

danloa:
Resolves #455 

- The Request Data option was added to the SelectList Inspector
- The source to be used should be an array in the request.

### To Test:
- Use  the master branch from  vue-form-elements


src/components/inspector/options-list.vue
```diff
@@ -31,7 +31,7 @@
         </div>
 
         <div class="card-footer text-right p-2">
-          <button type="button" class="btn btn-sm btn-outline-secondary mr-2" @click="showOptionCard=false">
+          <button type="button" class="btn btn-sm btn-outline-secondartr mr-2" @click="showOptionCard=false">
```
ryancooley: Should be `secondary`?

src/components/inspector/options-list.vue
```diff
@@ -329,6 +339,10 @@ export default {
       return '';
     },
     dataObjectOptions() {
+      if (!this.dataName) {
+        console.log('antes de');
```
ryancooley: Remove `console.log`

Issue #498, opened 2020-01-14
Date Formatting Adding A Day


velkymx:
https://www.dropbox.com/s/gb2clprswiitrz8/date-adding-one-date.mp4?dl=0

PR #502, opened 2020-01-16
Use timezone for setting default date value

chipit24:
This PR, along with https://github.com/ProcessMaker/vue-form-elements/pull/153, fixes https://github.com/ProcessMaker/screen-builder/issues/498.

This PR ensures that the default value for the datepicker is set using the user's timezone, and sets it to the start of the day if the "date" data type for the control is selected.

### DIFF
--- a/src/itemProcessingUtils.js
+++ b/src/itemProcessingUtils.js
@@ -1,3 +1,5 @@
+import moment from 'moment-timezone';
+
 function processFormItem(item) {
   if (item.component !== 'FormMultiColumn') {
     return item;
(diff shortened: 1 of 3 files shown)


src/itemProcessingUtils.js
```diff
@@ -28,8 +30,24 @@ export function getDefaultValueForItem(item) {
...
   }
+
   return defaultValue;
 }
+
+function generateNewDate(dataFormat) {
+  let timezone = moment.tz.guess();
+
+  if (typeof ProcessMaker !== 'undefined' && ProcessMaker.user && ProcessMaker.user.timzone) {
+    timezone = ProcessMaker.user.timzone;
```
lisa-fehr:
```suggestion
    timezone = ProcessMaker.user.timezone;
```

PR #506, opened 2020-01-17
Add missing datetime format

### LINKED ISSUE #505: Add missing time mask
Relates to https://github.com/ProcessMaker/screen-builder/issues/499

It is picking the default format for m/d/Y H:i

Screen Shot 2020-01-17 at 11 10 35 AM
Screen Shot 2020-01-17 at 11 10 54 AM
Screen Shot 2020-01-17 at 11 13 23 AM

lisa-fehr:
Closes #505

### DIFF
--- a/src/components/renderer/form-masked-input.vue
+++ b/src/components/renderer/form-masked-input.vue
@@ -57,22 +57,19 @@ const componentTypesConfigs = {
   percentage: 'getPercentageFormat',
 };
 
-function defaultMask()
-{
-  return {
-    date: ['##/##/####'],
-    dateTime: ['##/##/#### #:## SS', '##/##/#### ##:## SS'],
-  };
-}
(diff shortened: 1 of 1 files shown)


src/components/renderer/form-masked-input.vue
```diff
@@ -116,7 +119,7 @@ export default {
     getMask() {
       const format = this.dataFormat === 'date' ? this.getUserDateFormat() : this.getUserDateTimeFormat();
       return typeof masks[format] === 'undefined'
-        ? masks['default']
+        ? defaultMask()
```
chipit24: Why does `defaultMask` need to be a function? Is it because the mask is modified and so a new object reference is required in each case?

lisa-fehr: Fair point

PR #518, opened 2020-01-23
Nested Screen implementation

caleeli:
This PR includes:

- A control to nest a screen
- Includes an optional data variable. If it is blank, it will reference the top-level of the data object.

image

image


src/components/renderer/form-nested-screen.vue
```diff
@@ -0,0 +1,95 @@
...
+  <vue-form-renderer
+    class="form-nested-screen"
+    :placeholder="placeholder"
+    v-model="data"
+    :config="config"
+    mode="preview"
+    :computed="computed"
+    :custom-css="customCSS"
+    :watchers="watchers"
+    v-on:css-errors="cssErrors = $event"
```
chipit24: In almost all places in the app, we use the `@` shorthand for attaching event listeners. Can you change this to:
```suggestion
    @css-errors="cssErrors = $event"
```

src/mixins/multiselectApi.js
```diff
@@ -0,0 +1,64 @@
...
+      },
+    },
+  },
+  methods: {
+    change(value) {
+      this.$emit('input', this.storeId ? get(value, this.trackBy) : value);
+    },
+    loadOptions(filter) {
+      window.ProcessMaker.apiClient
+        .get(this.api + (typeof filter === 'string' ? '?filter=' + filter : ''))
```
chipit24: `window.ProcessMaker.apiClient` is an instance of [`axios`](https://github.com/axios/axios), so you can speicfy the query string (`?` part) of the URL through an options object:
```suggestion
        .get(this.api, { params: { filter } })
```

src/mixins/multiselectApi.js
```diff
@@ -0,0 +1,64 @@
...
+        .then(response => {
+          this.options = response.data.data || [];
+        });
+    },
+    loadSelected(value) {
+      window.ProcessMaker.apiClient
+        .get(this.api + '/' + value)
```
chipit24: This might read better as a template string:
```suggestion
        .get(`${this.api}/${value}`)
```

tests/e2e/specs/NestedScreen.js
```diff
@@ -0,0 +1,32 @@
+describe('Test Nested screen', () => {
```
chipit24: Having the word "Test" in your test title is redundant. It's already clear that this is a test. Also, the other test file is named with a `.spec.js` suffix. Can you rename this file to `NestedScreen.spec.js`?

tests/e2e/specs/NestedScreen.js
```diff
@@ -0,0 +1,32 @@
...
+  it('With a variable defined', () => {
+    cy.visit('/');
+    cy.get('#controls > div:contains("Nested Screen")').drag('#screen-container div:contains("Drag an element here")', 'bottom');
```
chipit24: Consider using more declarative selectors in your tests. Tags, class names and IDs can change any time and cause your tests to fail even though the functionality hasn't changed. Have a read through https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements.

tests/e2e/specs/NestedScreen.js
caleeli: Fixed

tests/e2e/specs/NestedScreen.js
caleeli: Fixed

caleeli:
The observations were fixed

PR #544, opened 2020-02-17
Enhance Button with Toggle Value Functionality (option for submit / do nothing) 

### LINKED ISSUE #543: Enhance Button with Toggle Value Functionality (option for submit / do nothing)
reference https://github.com/ProcessMaker/processmaker/issues/2808

marcoAntonioNina:
closes #543, 
Resolves https://github.com/ProcessMaker/processmaker/issues/2808

## Changes
- Add dropdown for button, types submit, script, default submit.
- add default value in button type script.
- button type script need config watcher
- value button type script is only a counter, to verify the data transition.

### DIFF
--- a/src/components/renderer/form-button.vue
+++ b/src/components/renderer/form-button.vue
@@ -19,6 +19,11 @@ export default {
       };
     },
   },
+  data() {
+    return {
+      countExecutionScript: 0,
+    }
+  },
   methods: {
     setValue(parent, name, value) {
       if (parent.items) {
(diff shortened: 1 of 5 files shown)


src/components/renderer/form-button.vue
```diff
@@ -28,6 +33,11 @@ export default {
       }
     },
     click() {
+      if (this.event === 'script') {
+        this.countExecutionScript++;
+        this.$emit('input', this.countExecutionScript);
```
chipit24: If execution enters this `if` block, then the component will end up emitting twice, once here, and once below where it calls `this.$emit(this.event, this.eventData)`. Is this intentional?

PR #586, opened 2020-02-26
Options props is deprecated

### LINKED ISSUE #558: Options props is deprecated
Deprecation warning for add sortable. Create/Edit a form-type screen. Open dev tools, go to the console

Options props is deprecated, add sortable options directly as vue.draggable item, or use v-bind. 

See https://github.com/SortableJS/Vue.Draggable/blob/master/documentation/migrate.md#options-props

marcoAntonioNina:
closes #558 

## Changes
- Component Draggable "props" deprecated, migrate to v-bind

### DIFF
--- a/src/components/editor/loop.vue
+++ b/src/components/editor/loop.vue
@@ -3,7 +3,7 @@
     <draggable
       style="min-height: 80px;"
       v-model="items"
-      v-bind="{group: {name: 'controls'}}"
+      group="controls"
     >
       <div class="control-item"
         :class="{selected: selected === element, hasError: hasError(element)}"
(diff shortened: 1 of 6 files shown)


src/components/editor/loop.vue
```diff
@@ -3,7 +3,7 @@
     <draggable
       style="min-height: 80px;"
       v-model="items"
-      :options="{group: {name: 'controls'}}"
+      v-bind="{group: {name: 'controls'}}"
```
chipit24: I think using a single prop reads better.
```suggestion
      group="controls"
```

Looking at the docs (https://github.com/SortableJS/Vue.Draggable/blob/master/documentation/migrate.md#options-props), it seems that the above should be valid.

marcoAntonioNina: the changes were made

PR #596, opened 2020-02-27
Change copy of "computed properties" to "calculated properties"

### LINKED ISSUE #594: Calculated Properties
Make sure that the term Calculated Properties is used everywhere. Currently it's presented as Computed.

johnblackspear:
This fixes #594.

Goes hand in hand with PR https://github.com/ProcessMaker/processmaker/pull/2904

### DIFF
--- a/src/components/computed-properties.vue
+++ b/src/components/computed-properties.vue
@@ -3,7 +3,7 @@
     ref="modal"
     size="lg"
     id="computed-properties"
-    :title="$t('Computed Properties')"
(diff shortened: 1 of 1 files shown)


src/components/computed-properties.vue
```diff
@@ -3,7 +3,7 @@
     ref="modal"
     size="lg"
     id="computed-properties"
```
tuantruong625: Looks like `id="computed-properties"` is being used or can be changed to `calculated-properties`.

PR #597, opened 2020-02-27
Nested submit button does nothing

### LINKED ISSUE #559: Nested submit button does nothing
**Steps to Reproduce:**

Have a screen with another screen nested in

Have the nested screen use a submit button

Run the screens

Press on the nested submit button

**Current Behavior:**

The button does nothing

Expected Behavior:

Remove / Hide submit button in the nested screen

https://www.dropbox.com/s/opzt2x35p5s5l5d/nestedSubmit.mp4?dl=0

marcoAntonioNina:
closes #559 

## Changes
- Hide submit buttons on nested screen.

### DIFF
--- a/src/components/renderer/form-nested-screen.vue
+++ b/src/components/renderer/form-nested-screen.vue
@@ -54,21 +54,24 @@ export default {
     },
   },
   methods: {
-    elementsValid(config) {
+    isSubmitButton(item) {
+      return item.config && item.component === 'FormButton' && item.config.event === 'submit';
+    },
+    hideSubmitButtons(config) {
       config.forEach(item => {
 
         //If the element has containers
         if (Array.isArray(item)) {
-          this.elementsValid(item);
+          this.hideSubmitButtons(item);
         }
 
         //If the element has items
         if (item.items) {
-          this.elementsValid(item.items);
+          this.hideSubmitButtons(item.items);
         }
 
         //hidden buttons
-        if (item.config && item.component === 'FormButton' && item.config.event === 'submit') {
+        if (this.isSubmitButton(item)) {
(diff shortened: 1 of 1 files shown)


src/components/renderer/form-nested-screen.vue
```diff
@@ -54,21 +54,41 @@ export default {
     },
   },
   methods: {
+    elementsValid(config) {
```
chipit24: The name `elementsValid` doesn't really communicate anything to me. Since this function is hiding all submit buttons, I think a name like `hideSubmitButtons` would communicate intent much better.

src/components/renderer/form-nested-screen.vue
```diff
@@ -54,21 +54,41 @@ export default {
...
+          this.elementsValid(item);
+        }
+
+        //If the element has items
+        if (item.items) {
+          this.elementsValid(item.items);
+        }
+
+        //hidden buttons
+        if (item.config && item.component === 'FormButton' && item.config.event === 'submit') {
```
chipit24: Instead of having a comment (which is confusing, since it's not explaining what this block is even doing, but simply stating "hidden buttons"), I think it would be much more readable to break out the conditional into a function. Something like:

```js
if (this.isSubmitButton(item)) {
  item.config.hidden = true;
}
```

src/components/renderer/form-nested-screen.vue
marcoAntonioNina: applied change

src/components/renderer/form-nested-screen.vue
marcoAntonioNina: applied change

PR #599, opened 2020-02-27
Fix buttons in loops

### LINKED ISSUE #592: Buttons Not Working In Loops
The 'Page Navigation',  'Submit' & 'Add Record' buttons do not work within a Loop component. When selected the page refreshes.

chipit24:
Fixes #592.

The fix for this is described at https://github.com/ProcessMaker/screen-builder/issues/592#issuecomment-592125989.

### DIFF
--- a/src/components/renderer/form-loop.vue
+++ b/src/components/renderer/form-loop.vue
@@ -34,7 +34,7 @@ export default {
       let items = this.items;
       return [{
         name: 'LoopItem',
-        items: items,
+        items,
       }];
     },
     times() {
@@ -52,12 +52,12 @@ export default {
 
       try {
         times = Mustache.render(times, this.transientData);
-      } catch (error) { }
-  
+      } catch (error) {}
+
       times = parseInt(times);
 
       if (Number.isNaN(times)) {
-        return []
+        return [];
       }
 
       if (times > 100) {
@@ -65,7 +65,7 @@ export default {
       }
 
       return [...Array(times).keys()];
-    }
+    },
   },
   watch: {
     transientData: {
@@ -93,7 +93,7 @@ export default {
     },
     times() {
       this.setupMatrix();
-    }
+    },
   },
   methods: {
     setMatrixValue(i, v) {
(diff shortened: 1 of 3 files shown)


src/components/renderer/form-loop.vue
```diff
@@ -1,18 +1,16 @@
 <template>
   <div class="form-group">
-    <div v-for="loopIndex in times" :key="loopIndex">
-      <form>
```
chipit24: This `form` element was the cause of the issue. Because nested buttons didn't specify a `type`, they defaulted to `"submit"`, causing this form to submit and reload the page.

I also moved the `v-for` from the `div` to `vue-form-renderer` and removed that redundant `div`.

src/components/renderer/form-loop.vue
```diff
@@ -1,18 +1,16 @@
...
+      v-for="loopIndex in times"
+      :key="loopIndex"
+      :data="getMatrixValue(loopIndex)"
+      :config="rendererConfig"
+      :computed="null"
+      :custom-css="null"
+      :watchers="null"
+      @submit="submit"
+      @update="setMatrixValue(loopIndex, $event)"
+    />
```
chipit24: I didn't change any of the props here, I re-arranged them to group bindings/props data and events.

sanjacornelius:
@chipit24  Submit & the Record List 'Add Record' button works but the 'Page Navigation' button is still having issues. The button no longer refreshes the page but now it doesn't load the configured page.

<a href="https://www.loom.com/share/30b1ef2989434928b595c9c711499543"> <p>27 February, 2020 - Loom Recording - Watch Video</p>  </a>

nolanpro:
@chipit24 individual iterations were in their own `<form>` element because that was the only way I could isolate radio groups. Do you know a better way? Try adding a select list and set it to render as radio buttons. Make sure they can operate independently in each loop iteration.

chipit24:
@sanjacornelius I swear I tested with navigation buttons but I guess I missed that :O I'll look into what's going on there.

And thanks @nolanpro! I'll test with radio buttons and update my PR. I may be able to catch the submit in the form and `preventDefault`.

chipit24:
@sanjacornelius @nolanpro Radio and navigation buttons should now be fixed.

PR #604, opened 2020-02-28
Record List File Uploads cannot be downloaded

### LINKED ISSUE #548: Record List File Uploads cannot be downloaded
In record list I can upload files but when I use the download file in the next task to download I can only view the name and I can't download

[1Record List File Upload Test.json.txt](https://github.com/ProcessMaker/processmaker/files/4243097/1Record.List.File.Upload.Test.json.txt)

The files are being properly seen in the "files" tab when looking at the request, but when opening the screen associated to the task, the File Download control does not seem to work.

Here's a quick example of the data being sent with the request:

```
"file_upload": "IMG_4899.png",
    "form_text_1": [
        {
            "record_list_file_upload": "what.png"
        }
    ],
    "form_button_1": null,
    "record_list_file_upload": null
```

danloa:
Resolves #548 

- Files are downloaded if the displayed column in the RecordList is associated with an uploaded file.
- To test use a RecordList that isn't embedded within a MultiColum control because there is a separate issue when using RecordLists and MultiColumns.

### DIFF
--- a/src/components/renderer/form-record-list.vue
+++ b/src/components/renderer/form-record-list.vue
@@ -40,6 +40,9 @@
         <template slot="mustache" slot-scope="{rowData, rowField}">
           {{ mustache(rowField, rowData) }}
         </template>
+        <template slot="filedownload" slot-scope="{rowData, rowField}">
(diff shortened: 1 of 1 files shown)


danloa:
@velkymx Conflicts were resolved

PR #742, opened 2020-05-20
Fix validation select value

### LINKED ISSUE #691: Validation Edit Missing Action Buttons
While I know that changing the values will automatically update, it's confusing to users that there is no way to save and close the dialogue. 

image

Add buttons to the bottom of the dialogue such as these:

image

Clicking update closes the forms and returns to the validation rules list:

image

marcoAntonioNina:
Fixed #691 
disable button save

### DIFF
--- a/src/components/inspector/validation-select.vue
+++ b/src/components/inspector/validation-select.vue
@@ -302,16 +302,15 @@ export default {
     selectedOption: {
       deep: true,
       handler(value) {
+        this.optionError = '';
+        this.disableBtn = true;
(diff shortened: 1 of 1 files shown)


sanjacornelius:
@marcoAntonioNina  I have a PR for this issue that uses less code manipulation. The issue was due to the multiselect tracking the 'value' which caused similar option values to be preselected giving them the bold styling within the dropdown. 
https://github.com/ProcessMaker/screen-builder/pull/743/files

PR #792, opened 2020-08-13
Do not emit if parent var does not exist yet

nolanpro:
Fixes https://processmaker.atlassian.net/browse/FOUR-1742

The console errors were shown because a watcher was run when the component is mounted that emits the _parent variable,  but the parent variable was still undefined. The _parent variable is what allows a record to read and modify the main screen's data from within a record list entry.

This was showing a console error but did not effect functionality.

The fix was to check that the _parent variable exists before emitting it.


caleeli:
This issue was replicated, and verified the PR fixes the error.

image

PR #824, opened 2020-11-10
Fix screen freezing when calculated property returns an array

nolanpro:
Fixes https://processmaker.atlassian.net/browse/FOUR-2148

The watcher was being re-triggered when a value was set even if it was the same. This was only effecting arrays, not single values. Added logic to check if the values are the same.

### DIFF
--- a/src/mixins/ScreenBase.js
+++ b/src/mixins/ScreenBase.js
@@ -33,10 +33,37 @@ export default {
       if (object && value !== undefined) {
         const splittedName = name.split('.');
         splittedName.forEach((attr, index) => {
-          const setValue = index < splittedName.length - 1 ? get(object, attr) || get(defaults, attr) || {} : value;
-          if (isEqual(setValue, get(object, attr))) {
+
+          let isLastElement, setValue;
+          const originalValue = get(object, attr);
+
+          if (index === splittedName.length - 1) {
+            isLastElement = true;
+          } else {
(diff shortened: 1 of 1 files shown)


src/mixins/ScreenBase.js
```diff
@@ -33,10 +33,14 @@ export default {
       if (object && value !== undefined) {
         const splittedName = name.split('.');
         splittedName.forEach((attr, index) => {
+          const setValue = index < splittedName.length - 1 ? get(object, attr) || get(defaults, attr) || {} : value;
```
velkymx, line 36: Please format and remove shorthand

caleeli:
Hi yes the code was shorthanded, I think this could be more understandable:
```
    setValue(name, value, object = this, defaults = object) {
      if (object && value !== undefined) {
        const currentValue = get(object, name);
        // Skip if equal
        if (_.isEqual(currentValue, value)) {
            return;
        }
        // Prepare base object if not defined
        let base = _.defaultsDeep(object, defaults);
        const attrs = name.split('.');
        const lastAttr = attrs.pop();
        base = attrs.reduce((base, attr) => {
            let value = get(base, attr);
            if (value === undefined) {
                value = {};
                this.$set(base, attr, value);
            }
            return value;
        }, base);
        // Set value
        this.$set(base, lastAttr, value);
      }
    },

```

PR #851, opened 2020-12-02
Setup watchers after the component mounts so defaults are already set

nolanpro:
Fixes https://processmaker.atlassian.net/browse/FOUR-2367

The watchers were being set up before the defaults were set. This caused the watcher to fire when the defaults were set. Since theres no guarantee what order the extensions are run in, this PR sets up the watchers to listen for changes after the component is mounted (after all the extensions have run)

### DIFF
--- a/src/mixins/extensions/Watchers.js
+++ b/src/mixins/extensions/Watchers.js
@@ -13,7 +13,7 @@ export default {
       if (definition.watchers) {
         screen.mixins.push(watchersMixin);
         definition.watchers.forEach((watcher) => {
-          this.addWatch(screen, watcher.watching, `this.queueWatcher(${JSON.stringify(watcher)});`);
+          this.addMounted(screen, 'var that = this; this.$watch("' + watcher.watching + '", function() { that.queueWatcher(' + JSON.stringify(watcher) + '); });');
         });
       }
     },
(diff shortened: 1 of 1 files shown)


src/mixins/extensions/Watchers.js
```diff
@@ -13,7 +13,7 @@ export default {
       if (definition.watchers) {
         screen.mixins.push(watchersMixin);
         definition.watchers.forEach((watcher) => {
-          this.addWatch(screen, watcher.watching, `this.queueWatcher(${JSON.stringify(watcher)});`);
+          this.addMounted(screen, 'var that = this; this.$watch("' + watcher.watching + '", function() { that.queueWatcher(' + JSON.stringify(watcher) + '); });');
```
caleeli, line 16: Could you try Arrow Function declaration instead the use of `var that = this;`:
```
    this.addMounted(
        screen,
        'this.$watch("' + watcher.watching + '", () => { this.queueWatcher(' + JSON.stringify(watcher) + '); });'
    );
```

caleeli (changes requested):

src/mixins/extensions/Watchers.js
nolanpro: @caleeli The problem is it's a string so Babel doesn't convert it to ES5

caleeli (approved):

boliviacoca (approved):

PR #909, opened 2021-02-01
Add Custom Formatting Support

sanjacornelius:
**Jira Ticket**
https://processmaker.atlassian.net/browse/FOUR-2740

<h2>Change</h2>

Within the 'Advanced' accordion for a Line Input control, there is now an area to allow custom mask formatting. 

Screen Shot 2021-02-01 at 11 38 52 AM

**Video**

https://www.loom.com/share/bd9e8c2e926a4b3c85f810a11c4f12b4

### DIFF
--- a/src/components/renderer/form-masked-input.vue
+++ b/src/components/renderer/form-masked-input.vue
@@ -47,12 +47,14 @@ const componentTypes = {
   date: 'the-mask',
   datetime: 'the-mask',
   percentage: 'inputmasked',
+  custom: 'the-mask'
 };
 const componentTypesConfigs = {
   currency: 'getCurrencyFormat',
   date: 'getDateFormat',
   datetime: 'getDatetimeFormat',
   percentage: 'getPercentageFormat',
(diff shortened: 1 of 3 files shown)


src/components/renderer/form-masked-input.vue
```diff
@@ -76,38 +78,44 @@ export default {
...
-          break;
-        case 'array':
-          break;
-        default:
-          newValue = newValue.toString();
-          break;
+      if (this.customFormatter) {
+        newValue = newValue.replace(/[^\w]/g, '');
+        return newValue;
+      } else {
```
danloa, line 84: As there is a return command inside the if-block, to have an else-block  is not necessary. Furthermore, we'll reduce the code nesting

sanjacornelius: Good Catch! I'll fix it!

danloa (approved):

boliviacoca (approved):

PR #943, opened 2021-03-02
Date entered with the Date Picker cannot be submitted with the Submit button

marcoAntonioNina:
Fixed [FOUR-2892](https://processmaker.atlassian.net/browse/FOUR-2892)

- Remove custom validation before

### DIFF
--- a/src/components/vue-form-renderer.vue
+++ b/src/components/vue-form-renderer.vue
@@ -143,7 +143,9 @@ export default {
       return child.errors();
     },
     isValid() {
-      //deprecated
+      /**
+       * @deprecated Since version 2.4.4
+       */
       const items = getItemsFromConfig(this.definition.config);
       let config = _.cloneDeep(this.definition.config);
 
--- a/src/components/vue-form-renderer.vue
+++ b/src/components/vue-form-renderer.vue
@@ -142,10 +142,12 @@ export default {
 
       return child.errors();
     },
+    /**
+     * Check if a form is valid for submitting
+     *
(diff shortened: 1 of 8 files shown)


src/components/vue-form-renderer.vue
```diff
@@ -143,6 +143,7 @@ export default {
       return child.errors();
     },
     isValid() {
```
caleeli, line 145: According to [JSDoc](https://en.wikipedia.org/wiki/JSDoc#JSDoc_tags) a function can be marked as deprecated this way:
```
/**
 * @deprecated Since version 2.4.4
 */
```

caleeli (changes requested):

src/components/vue-form-renderer.vue
marcoAntonioNina: changed

caleeli: docblock must go before function declaration, like: 
```
  /**
   * Calculates the circumference of the Circle.
   *
   * @deprecated since 1.1.0; use getCircumference instead
   * @return {number} The circumference of the circle.
   */
  calculateCircumference() {
    return 2 * Math.PI * this.radius
  }
```

caleeli (approved):

boliviacoca (approved):

nolanpro:
@marcoAntonioNina Looks like validation is not working when in a nested screen in a multi column

Kapture 2021-03-08 at 14 01 32

PR #957, opened 2021-03-23
Fix value path validation

caleeli:
Resolves: https://processmaker.atlassian.net/browse/FOUR-2876

Fix path validation when variable name includes dots like `result.name`


GooseCoder (approved):
lgtm

boliviacoca (approved):

PR #988, opened 2021-05-12
FOUR-3288: Watchers on Display Screens

danloa:
Fixes [https://processmaker.atlassian.net/browse/FOUR-3288](https://processmaker.atlassian.net/browse/FOUR-3288)

As before,  the select list shows all the variables of the current screen but additionally the user can type any string so that he can use any request variable name.

### DIFF
--- a/src/components/watchers-form.vue
+++ b/src/components/watchers-form.vue
@@ -36,6 +36,7 @@
               :helper="$t('Select the variable to watch on this screen or type any request variable name')"
               @open="loadVariables"
               @tag="addTag"
+              :tag-placeholder="$t('Press enter to use this variable')"
               data-cy="watchers-watcher-variable"
             />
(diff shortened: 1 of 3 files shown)


src/components/watchers-form.vue
```diff
@@ -36,6 +36,7 @@
               :helper="$t('Select the variable to watch on this screen or type any request variable name')"
               @open="loadVariables"
               @tag="addTag"
+              :tag-placeholder="$t('Press enter to use this variable')"
```
ryancooley, line 39: I added this line to change the message from "Press enter to create a tag" to "Press enter to use this variable".

Before:
Screen Shot 2021-05-12 at 10 04 04 AM

After:
Screen Shot 2021-05-12 at 10 04 43 AM

ryancooley (approved):

boliviacoca (approved):

PR #1007, opened 2021-06-01
It is not possible to create a variables Object with low bar "_"

marcoAntonioNina:
Fixes [FOUR-3477](https://processmaker.atlassian.net/browse/FOUR-3477)

- Change regex validation variable name

### DIFF
--- a/src/form-control-common-properties.js
+++ b/src/form-control-common-properties.js
@@ -95,7 +95,7 @@ export const keyNameProperty = {
   config: {
     label: 'Variable Name',
     name: 'Variable Name',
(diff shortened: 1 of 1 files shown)


caleeli (changes requested):
This regex could be improved a little more, this is an invalid variable reference `foo_..bar`:
```
"foo_..bar".match(/^(?:[A-Za-z])(?:[0-9A-Z_.a-z])*[^.]$/)
```
but passes the validation.

marcoAntonioNina:
regex was changed.
the allowed names are as for example:

foo.bar
foo._bar
foo._bar.10
foo._bar.10.tes

caleeli (approved):

boliviacoca (approved):

PR #1026, opened 2021-07-28
FOUR-3575:  returned data is mixed when the watcher runs at the same time in two different sessions

danloa:
Fixes [https://processmaker.atlassian.net/browse/FOUR-3575](https://processmaker.atlassian.net/browse/FOUR-3575)

### DIFF
--- a/src/mixins/watchers.js
+++ b/src/mixins/watchers.js
@@ -12,6 +12,9 @@ export default {
   },
   methods: {
     queueWatcherSync(watcher) {
+      //add a random sufix to uniquely identify this watcher call
+      watcher.uid = watcher.uid + (Math.random().toString(36)+'00000000000000000').slice(2, 16) + Date.now();
+
       if (watcher.synchronous) {
         // lock screen with watcher's popup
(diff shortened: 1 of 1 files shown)


src/mixins/watchers.js
```diff
@@ -12,6 +12,9 @@ export default {
   },
   methods: {
     queueWatcherSync(watcher) {
+      //add a random sufix to uniquely identify this watcher call
```
caleeli, line 15: Could you try https://www.npmjs.com/package/uuid for a more robust solution?

caleeli (commented):
Could you try https://www.npmjs.com/package/uuid for a more robust solution?

src/mixins/watchers.js
danloa: The unique ui is per watcher and time (milliseconds) so it seems a good solution in this context, in that way we avoid to use an extra module for just this.

caleeli (approved):

boliviacoca (approved):

PR #1031, opened 2021-08-05
Upgrade inputmask package

nolanpro:
The old version allowed the user to put the cursor at the end and start typing, resulting in wrong comma placement:
Kapture 2021-08-05 at 11 31 18

Upgrading forces the user to put the cursor in the correct position


caleeli (approved):
Very nice

boliviacoca (approved):

PR #1054, opened 2021-09-23
Ticket #742 Validation in Hidden Elements (for 4.2)

eiresendez:
- Changes: A condition was added to not validate the hidden elements.
- Fixes:  http://tickets.pm4overflow.com/tickets/742


danloa (changes requested):
The validations does not respond when the a control changes from  hidden to visible
Peek 2021-09-23 15-54

PR #1067, opened 2021-10-05
Bugfix Attribute name same as the Variable name

caleeli:
Issue: When a field name is set as for example: `address.address.city` it breaks the screen builder.

https://user-images.githubusercontent.com/8028650/136061824-14866710-c145-4c93-9685-cebe5254886e.mp4

This PR includes:

- Test to replicate the issue
- Fix to solve the problem

https://user-images.githubusercontent.com/8028650/136061905-762922eb-57ae-4f1a-bd13-a9993d988fe4.mp4

### DIFF
--- a/src/mixins/ScreenBase.js
+++ b/src/mixins/ScreenBase.js
@@ -117,6 +117,7 @@ export default {
             setValue
           );
           object = get(object, attr);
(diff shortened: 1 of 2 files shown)


eiresendez (approved):
It works as expected!

PR #1074, opened 2021-10-07
Fix Required IF rule with checkbox field

caleeli:
Fixes: http://tickets.pm4overflow.com/tickets/1005

The Required IF/Required Unless does not work with boolean values.

This PR includes:

- Test to reqplicate the issue with Required IF
- Test to reqplicate the issue with Required Unless
- Fix validation rules


eiresendez (approved):
It works correctly when variable value is boolean

PR #1086, opened 2021-10-14
Hide Tooltips for Submit buttons

sanjacornelius:
Ticket [1060](http://tickets.pm4overflow.com/tickets/1060)

<h2>Changes</h2>

- Removes the Tooltip configs in the inspector panel for the Submit type buttons.

**Submit Button**

Screen Shot 2021-10-14 at 3 13 25 PM

**Regular Button**

Screen Shot 2021-10-14 at 3 13 47 PM

### DIFF
--- a/src/components/inspector/tooltip.vue
+++ b/src/components/inspector/tooltip.vue
@@ -34,6 +34,7 @@
 </template>
 
 <script>
+import _ from 'lodash';
 
 export default {
   inheritAttrs: false,


src/components/inspector/tooltip.vue
```diff
@@ -129,6 +129,11 @@ export default {
       variant: '',
     };
   },
+  computed: {
+    buttonType() {
+      return this.$attrs.selectedControl.config.event;
```
ryancooley, line 134: Is there ever a case where part of this entire path (`this.$attrs.selectedControl.config.event`) may cease to exist and then return an undefined error in the console? Just want to make sure we have some error handling here. Might be good to use the [Lodash _.get method](https://lodash.com/docs/#get) here to prevent any errors in the future.

ryancooley: Something similar to `_.get(this.$attrs, 'selectedControl.config.event')` should work.

ryancooley (approved):

PR #1108, opened 2021-11-11
Bugfix/FOUR-4329: Fixed watcher triggering on a second page with checkbox (FOR 4.2)

agustinbusso:
Ticket [FOUR-4329](https://processmaker.atlassian.net/browse/FOUR-4329)

Default checkbox value in second page was setted to null instead false, so when moving to second page, the value of the checkbox that initially was null changed to false and watcher was detecting that reactivity. Now checkbox is setted as false instead null.

**Working video**

https://user-images.githubusercontent.com/90727999/141355450-733395d1-9728-4c74-940c-a4373696c025.mov

### DIFF
--- a/src/mixins/extensions/DefaultValues.js
+++ b/src/mixins/extensions/DefaultValues.js
@@ -15,8 +15,8 @@ export default {
             this.setupDefaultValue(screen, name, `(function() {${config.defaultValue.value}}).bind(this.vdata)()`);
           }
         }
-        if (config.initiallyChecked) {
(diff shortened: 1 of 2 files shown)


caleeli:
@agustinbusso 
Please update the test `ComplexScreen.spec.js` considering your change.

image

caleeli (changes requested):
@agustinbusso 
Please update the test `ComplexScreen.spec.js` considering your change.

image

PR #1115, opened 2021-11-29
Test Component Select List

marcoAntonioNina:
- add test select list object
- remove field value in component select list when value returned is obejct.


caleeli (approved):
The following scenarios were tested in PM 4.1-develop:
- Select List with Data provided
- Select list with request data
- Select List with Objects
- Select List with data connector
- Select List with collections
- Select List Multi-Select
- Select List in Collection screen
- User Task -> Task
- WebEntry -> Task

PR #1126, opened 2021-12-15
Bugfix/FOUR-4903: Date Field validation not working with calc prop (For 4.2)

agustinbusso:
## Issue & Reproduction Steps
**Issue 1**
Validations not working for calculated properties (after date, before date, after or equal date, before or equal date)

**Issue 2**
Client is returning a date with the format DD/MM/YYYY and should be YYYY-MM-DD as described in the [documentation](https://processmaker.gitbook.io/processmaker/designing-processes/design-forms/screens-builder/control-descriptions/validation-rules-for-validation-control-settings#after-date)

**Issue 3**
If you have a datetime input with the value eg. 15/12/2021 11:52 and we want to validate if **after** (not after or equal) than 15/12/2021 the output should be false becase is not after, is equal date, but is returning true because the currentDate is 15/12/2021 00:00 and the input date is 15/12/2021 11:52. The same happens with the other cases.

Reproduction steps
- Have a screen with a date field
- Add a validation rule "Before or Equal to Date"
- Add a variable in mustache currentDate
- Have a calc prop to return the current date with JS to the variable currentDate

Sample calc property code (set name of the calc property to currentDate):
```
today = new Date();
var sp = '-';
var dd = today.getDate();
var mm = today.getMonth()+1; //As January is 0.
var yyyy = today.getFullYear();

if(dd<10) dd='0'+dd;
if(mm<10) mm='0'+mm;
// return (dd+sp+mm+sp+yyyy);  First issue detected is that client is returning a date in format DD/MM/YYYY
return (yyyy+sp+mm+sp+dd);
```
- Go to Preview and select a date Before or equal that currentDate calc property. You will note that validation is not working.

## Solution
- When doing moment(get(dataWithParent, before_or_equal, before_or_equal)) searching from before_or_equal index is not finding the key inside dataWithParent because it is passed as {{currentDate}} instead currentDate, so added a regex to remove mustaches {{ }}.
- Added endOf('day') to before_or_equal and after validations to work properly
- Added startOf('day') to after_or_equal and before validations to work properly 

## How to Test
- Have a screen with a date field
- Add a validation rule "Before or Equal to Date"
- Add a variable in mustache currentDate
- Have a calc prop to return the current date with JS to the variable currentDate
Use the previous sample calc property code and set name of the calc property to currentDate
- Go to Preview and select a date Before or equal that currentDate calc property. You will note that validation is working now. **Please test it with After date, After or equal date, before date and Before or equal date.**

**Working video**

https://user-images.githubusercontent.com/90727999/146244650-e4b6a95f-2829-462c-81ca-c9bdfa3c625d.mov

## Related Tickets & Packages
- [FOUR-4903](https://processmaker.atlassian.net/browse/FOUR-4903)

### DIFF
--- a/src/components/inspector/validation-select.vue
+++ b/src/components/inspector/validation-select.vue
@@ -57,9 +57,12 @@
             </div>
           </b-card-header>
           <b-collapse :id="formatRuleContentAsId(rule.content)" :accordion="formatRuleContentAsId(rule.content)" :visible="rule.visible" role="tabpanel">
-            <b-card-body> 
-              <div class="p-2"> 
+            <b-card-body>
+              <div class="p-2">
                 <div v-for="config in rule.configs" :key="config.label" class="mb-2">
+                  <div v-if="rule.showMustacheHelper">
+                    <mustache-helper/>
+                  </div>
                   <div v-if="config.type === 'FormInput'">
                     <form-input :label="config.label" :name="config.name || config.label" v-model="config.value" :validation="config.validation" :helper="config.helper"/>
                   </div>
@@ -92,6 +95,7 @@
 
 <script>
 import { FormInput } from '@processmaker/vue-form-elements';
+import MustacheHelper from './mustache-helper';
 import _ from 'lodash';
 import InputVariable from '../inspector/input-variable';
 
@@ -100,6 +104,7 @@ export default {
   components: {
     FormInput,
     InputVariable,
+    MustacheHelper,
   },
   data() {
     return {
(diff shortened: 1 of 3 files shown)


caleeli:
@agustinbusso the datetime validations are failing with your change:

With changes:

https://user-images.githubusercontent.com/8028650/146587347-d1f94111-2fc1-425a-bc47-04b7ea09f91d.mp4


Previous version:
https://user-images.githubusercontent.com/8028650/146587158-95bf1edd-4d2f-43fc-adef-e65ea1736311.mp4


Screen:
[FOUR-4903.zip](https://github.com/ProcessMaker/screen-builder/files/7736905/FOUR-4903.zip)

caleeli (changes requested):
Please review the datetime validations.

PR #1169, opened 2022-01-20
Bugfix/FOUR-5103: Using the sort option in a record list, doesn't update the records when using the edit action

agustinbusso:
## Issue & Reproduction Steps
When try to edit or delete a row in record list after sorting, it is taking the wrong record.
- Create a screen with only a record list control
- Add one column named ID to the record list
- Create a new page in order to add values to the record list
- Add a line input with ID as the variable
- In the first page check the Editable option and choose the new page in the record form field
- Click on preview add some rows
- Sort the column and try to edit one record or delete

The data shown in the edit option and in the record list is not the same. If you try to delete the record, will remove the wrong record

## Solution
- After sorting, listen the sortChanged event and sort also the parent tableData (to have the same order that the local vuetable) and retrieve the edit or delete row from the tableData variable

## How to Test
- Create a screen with only a record list control
- Add one column named ID to the record list
- Create a new page in order to add values to the record list
- Add a line input with ID as the variable
- In the first page check the Editable option and choose the new page in the record form field
- Click on preview add some rows
- Sort the column and try to edit one record or delete
- You will get the correct row for edit and delete.

Run tests/e2e/specs/RecordList.spec.js tests

**Working video**

https://user-images.githubusercontent.com/90727999/150342975-f4741998-f06f-4c67-afb3-d6a324f85484.mov

## Related Tickets & Packages
- [FOUR-5103](https://processmaker.atlassian.net/browse/FOUR-5103)

## Code Review Checklist

### DIFF
--- a/src/components/renderer/form-record-list.vue
+++ b/src/components/renderer/form-record-list.vue
@@ -24,6 +24,8 @@
         :css="css"
         :empty-text="$t('No Data Available')"
         :current-page="currentPage"
+        @sort-changed="sortChanged"
+        @input="onInput"
         data-cy="table"
       >
         <template #cell()="{index,field,item}">
@@ -242,6 +244,7 @@ export default {
         from,
         to: value.length,
         data: value,
+        lastSortConfig: false,
       };
       return data;
     },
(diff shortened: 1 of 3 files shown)


eiresendez (approved):
It works as expected! Tests are green.

https://user-images.githubusercontent.com/90741874/150371456-596d67b5-17fd-4495-98d7-4ff0ee531fc0.mov

ryancooley (approved):

PR #1186, opened 2022-02-17
FOUR-5348: 4.2.- Variable is not created if has default value dependent

danloa:
## Issue & Reproduction Steps

The problem was produced because for the default value was used a JavaScript expression that didn't return any value. By default JavaScript returns "undefined". 

Expected behavior: 
Even if the the expression used for the default value of a select list is undefined, its value should be able to be set.

Actual behavior: 
The select list variable with the undefined default value it is not updated when an item is selected.

## Solution
If the JavaScript function returns undefined, by default the variable is initialized with a null value.

## How to Test
Steps to Reproduce:

    Create Screen

    Add select list 1 with default values

     

[{"content":"1","value":"1"},{"content":"2","value":"2"}]

Put default value = 2

Add select list 2 with data connector

Put default value JS

 
if (this.form_select_list_2 != 2) {

    return null;

}

Press preview button

Select an option in the second select list and verify that this 2nd dropdown variable update its corresponding variable.

## Related Tickets & Packages
- [https://processmaker.atlassian.net/browse/FOUR-5348](https://processmaker.atlassian.net/browse/FOUR-5348)

## Code Review Checklist

### DIFF
--- a/src/mixins/ScreenBase.js
+++ b/src/mixins/ScreenBase.js
@@ -93,7 +93,8 @@ export default {
     },
     tryFormField(variableName, callback, defaultValue = null) {
       try {
-        return callback() || '';
+        let result = callback();
+        return (result === undefined) ? null : result;
       } catch (e) {
         set(this.$v, `${variableName}.$invalid`, true);
         set(this.$v, `${variableName}.invalid_default_value`, false);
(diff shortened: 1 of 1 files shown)


eiresendez (approved):
To test it's necessary to use the condition: `if (this.form_select_list_1 != 2))`. 

Working perfectly. Tests passed.

https://user-images.githubusercontent.com/90741874/154550122-10f7062b-25f6-4f2a-bccf-4f98158d3ac8.mov

ryancooley (approved):

PR #1194, opened 2022-03-14
Bugfix FOUR-4986 - File Upload control inside a loop in the record list issues

eiresendez:
## Issue & Reproduction Steps
Expected behavior: 
- The documents uploaded should be shown correctly on the File Upload UI.

Actual behavior: 
- When we upload documents inside a loop control that at the same time this loop is inside a record list, on the next task it will not be visible or the first document will be visible twice.

## Solution
The issue was related to:

1. The `query.options.row_id` property for new items in the loop was generated with NULL value. This was fixed from the `form-recordlist.vue` file. 
2. The `query.options.data_name` property was generated with incorrect values. This was fixed by commenting out the `checkIfInRecordList` function.

## How to Test
- Import the process `26955.json` attached in JIRA
- Run a new request
- Add new records in record list
- Upload documents on the loop control
    - You can override the same file
    - Delete an item from the loop
    - Add more items and files within the loop.
    - Add another record in Record List and repeat

The files must be saved and/or deleted correctly in `storage/app/public` backend. When submitting the task, everything should be displayed correctly and the files can be downloaded.

## Related Tickets & Packages
- [FOUR-4986](https://processmaker.atlassian.net/browse/FOUR-4986)

## Code Review Checklist

### DIFF
--- a/src/components/renderer/file-upload.vue
+++ b/src/components/renderer/file-upload.vue
@@ -86,8 +86,6 @@ export default {
   mounted() {
     this.$root.$on('set-upload-data-name',
       (recordList, index, id) => {
-        // eslint-disable-next-line no-console
-        console.log(recordList, index, id);
         this.listenRecordList(recordList, index, id);
       });
 
@@ -99,7 +97,7 @@ export default {
 
     this.removeDefaultClasses();
     
-    this.checkIfInRecordList();
+    // this.checkIfInRecordList();
 
     this.setPrefix();
     if (this.$refs['uploader']) {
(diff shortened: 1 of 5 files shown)


src/components/renderer/file-upload.vue
```diff
@@ -95,7 +97,7 @@ export default {
 
     this.removeDefaultClasses();
     
-    this.checkIfInRecordList();
+    // this.checkIfInRecordList();
```
sanjacornelius, line 100: @eiresendez Can you remove this and the code associated with this method since it is no longer being used?

eiresendez: sure, done! @sanjacornelius

danloa (approved):

sanjacornelius (approved):

ryancooley (approved):

PR #1199, opened 2022-04-05
Bugfix/FOUR-5103_d: Using the sort option in a record list, doesn't update the records when using the edit action

agustinbusso:
## Issue & Reproduction Steps

- Create a screen with only a record list control
- Add one column named input record list
- Create a new page in order to add values to the record list
- Add a line input with input as the variable
- Go to preview and add some data
**- Try with different combinations, changing pages and sorting in different ways. IMPORTANT: The original ticket was reopened because when adding empty records to the record list in the sorting column, it was not sorting and retrieving the correct record.** For texting correctly you can add 2 inputs in the record list (input 1 and input2) and fill them in a pattern to make easy the testing. For example (please note the dashes, that means leave the field empty):
```
Input 1               Input 2
B                     2
A                     1
D                     4
C                     3
E                     5
-                     7
F                     -
-                     6
```
so you can know in an easy way if the record you clicked for update is the correct and also if after save the correct record was edited/deleted

Actual behavior: 
In some scenarios (For example when you add empty values in the record list), sorting was not was working correctly, and the editing/deleting element was wrongly selected.

## Solution
- Removed custom sorting function, and get the record by the unique row_id identifier

## How to Test
Test the steps above in reproduction steps
Run e2e test cases fr RecordList

**Working video**

https://user-images.githubusercontent.com/90727999/161844246-dbdb2dd3-1b6c-41fb-a365-e154505f7150.mov

## Related Tickets & Packages
- [FOUR-5103](https://processmaker.atlassian.net/browse/FOUR-5103)

## Code Review Checklist

### DIFF
--- a/src/components/renderer/form-record-list.vue
+++ b/src/components/renderer/form-record-list.vue
@@ -21,12 +21,11 @@
         :fields="tableFields"
         :items="tableData.data"
         :sort-compare-options="{ numeric: false }"
+        :sort-null-last="true"
         sort-icon-left
         :css="css"
         :empty-text="$t('No Data Available')"
         :current-page="currentPage"
-        @sort-changed="sortChanged"
-        @input="onInput"
         data-cy="table"
       >
         <template #cell()="{index,field,item}">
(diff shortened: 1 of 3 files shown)


eiresendez (approved):
It works properly when editing, deleting and sorting. Pagination also works as expected. Tested manually - ran e2e tests.

ryancooley (approved):

PR #1254, opened 2022-09-14
Added a check in the definition watcher to check if the defition.conf…

josechirivella:
Added a check in the definition watcher to check if the definition.config if empty. If empty, do nothing, if not, rebuild the screen

## Issue & Reproduction Steps

Expected behavior: 
`rebuildScreen` shouldn't be executed all the time unless needed.

Actual behavior: 
`rebuildScreen` is being executed in the most absurd scenarios. When you move line inputs from the order in the ScreenBuilder Designer, or switch from Designer to Preview.

## Solution
- Added a check using lodash to see if `definition.config` was empty since `config` contains mainly the configuration of the screen. If the config is populated, means that the screen has some configurations or something of means to rebuild the screen 👀 

## How to Test
Test the steps above
* Add a `console.count("Rebuild Screen method");` between 76 and 77 to check if the `rebuildScreen` method is being called

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-6740

## Code Review Checklist

### DIFF
--- a/src/components/screen-renderer.vue
+++ b/src/components/screen-renderer.vue
@@ -1,56 +1,86 @@
 <template>
   <b-container fluid>
-    <div v-if="displayAsyncLoading" ref="watchersAsynchronous" class="alert alert-light" role="alert">
-      <svg class="lds-gear" width="20px" height="20px" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 100 100" preserveAspectRatio="xMidYMid">
+    <div
+      v-if="displayAsyncLoading"
+      ref="watchersAsynchronous"
+      class="alert alert-light"
+      role="alert"
+    >
+      <svg
+        class="lds-gear"
+        width="20px"
+        height="20px"
+        xmlns="http://www.w3.org/2000/svg"
+        xmlns:xlink="http://www.w3.org/1999/xlink"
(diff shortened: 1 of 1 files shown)


src/components/screen-renderer.vue
```diff
@@ -1,56 +1,86 @@
...
-  name: 'screen-renderer',
+  name: "ScreenRenderer",
   components: { WatchersSynchronous, ScreenRendererError },
-  mixins: [ Json2Vue, CurrentPageProperty ],
+  mixins: [Json2Vue, CurrentPageProperty],
   data() {
     return {
       currentDefinition: null,
-      codigo: '',
-      self: this,
```
josechirivella, line 39: removed `self` since 1. It wasn't being used and 2. goes agaisn't some javascript rules such as 
```var this = that`

File: src/components/screen-renderer.vue [hunk 2]
```diff
@@ -1,56 +1,86 @@
...
   },
-  mounted() {
-    this.currentDefinition = cloneDeep(this.definition);
-    this.component = this.buildComponent(this.currentDefinition);
-  },
   watch: {
     definition: {
       deep: true,
       handler(definition) {
+        if (isEmpty(definition.config)) return;
```
josechirivella, line 76: main change

caleeli (approved):

PR #1255, opened 2022-09-16
Improve refresh validation event triggering to perform

caleeli:
## Issue & Reproduction Steps
For a form with validation rules and visibility rules combined, when an input text changes it triggers multiple times the `refresh-validation-rules` event causing performance problems.

## Expected behavior: 
The refresh of validation rules should be "debounced" to prevent it triggers multiple times on an input change.

## Actual behavior: 
The refresh of validation rules is triggered multiple times, it increases when the screen has many nested screens and loops.

## Solution
- Move the refresh validation rules logic to $store so this way it will be unique for the screen.
- Add refresh validation rules logic to submit button to make sure all the validations are updated before submit.
- Debounce the refresh validation rules function.

## How to Test
- In a Screen with visibility rules combined with validation rules
- Fill the values
- Check the validation is working as expected.

[Test Screen.zip](https://github.com/ProcessMaker/screen-builder/files/9585360/Test.Screen.zip)

https://user-images.githubusercontent.com/8028650/190676175-5516eaa7-a747-469c-ab5d-0a7dba97c11c.mp4

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-6719

## Code Review Checklist

### DIFF
--- a/src/components/renderer/form-button.vue
+++ b/src/components/renderer/form-button.vue
@@ -9,19 +9,18 @@
 <script>
 import Mustache from 'mustache';
 import { getValidPath } from '@/mixins';
+import { mapState } from 'vuex';
 
 export default {
   mixins: [getValidPath],
   props: ['variant', 'label', 'event', 'eventData', 'name', 'fieldValue', 'value', 'tooltip', 'transientData'],
   watch: {
-    "$store.state.globalErrorsModule": {
-      deep: true,
-      handler(validate) {
-        this.isInvalid = !validate.valid;
-      },
-    },
+    valid(valid) {
+      this.isInvalid = !valid;
+    }
   },
   computed: {
+    ...mapState('globalErrorsModule', ['valid']),
     classList() {
       let variant = this.variant || 'primary';
       return {
--- a/src/components/renderer/form-button.vue
+++ b/src/components/renderer/form-button.vue
@@ -9,40 +9,25 @@
 <script>
 import Mustache from 'mustache';
 import { getValidPath } from '@/mixins';
-import { isProxy } from 'is-proxy';
 
 export default {
   mixins: [getValidPath],
   props: ['variant', 'label', 'event', 'eventData', 'name', 'fieldValue', 'value', 'tooltip', 'transientData'],
   watch: {
-    '$attrs.validate': {
+    "$store.state.globalErrorsModule": {
       deep: true,
       handler(validate) {
-        if (validate && !isProxy(validate.vdata.$model)) {
-          this.errors = 0;
-          let message = '';
-          if (validate.$invalid) {
-            this.countErrors(validate.vdata);
-            this.countErrors(validate.schema);
-            message = this.errors === 1
-              ? 'There is a validation error in your form.'
-              : 'There are {{items}} validation errors in your form.';
-            message = this.$t(message, {items: this.errors});
-          }
-          this.$store.commit('globalErrorsModule/basic', {key: 'valid', value: !validate.$invalid});
-          this.$store.commit('globalErrorsModule/basic', {key: 'message', value: message});
-        }
+        this.isInvalid = !validate.valid;
       },
     },
   },
   computed: {
     classList() {
       let variant = this.variant || 'primary';
-      let isInvalid = this.$store.getters['globalErrorsModule/isValidScreen'] === false;
       return {
         btn: true,
         ['btn-' + variant]: true,
-        disabled: this.event === 'submit' && isInvalid,
+        disabled: this.event === 'submit' && this.isInvalid,
       };
     },
     options() {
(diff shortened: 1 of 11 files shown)


src/mixins/Json2Vue.js
```diff
@@ -2,7 +2,7 @@ import extensions from './extensions';
 import ScreenBase from './ScreenBase';
 import CountElements from '../CountElements';
 import ValidationsFactory from '../ValidationsFactory';
-import _, { debounce, isEqual } from 'lodash';
+import _, { isEqual } from 'lodash';
```
josechirivella, line 5: if you can, do a find a replace of all the `_.` and add the imports of the properties we are using from lodash

josechirivella: not a blocker

src/components/vue-form-renderer.vue
```diff
@@ -120,6 +127,9 @@ export default {
     this.scrollable = Scrollparent(this.$el);
   },
   methods: {
+    getMainScreen() {
+      return this.$refs.renderer && this.$refs.renderer.$refs.component
```
josechirivella, line 131: missing a semicolon 
```suggestion
      return this.$refs.renderer && this.$refs.renderer.$refs.component;
```

src/components/renderer/form-button.vue
```diff
@@ -9,40 +9,25 @@
...
 import Mustache from 'mustache';
 import { getValidPath } from '@/mixins';
-import { isProxy } from 'is-proxy';
 
 export default {
   mixins: [getValidPath],
   props: ['variant', 'label', 'event', 'eventData', 'name', 'fieldValue', 'value', 'tooltip', 'transientData'],
   watch: {
-    '$attrs.validate': {
+    "$store.state.globalErrorsModule": {
```
josechirivella, line 17: So. We have here multiple unnecessary watchers. One, which is watching the main default which is `globalErrorsModule`, the second one which is watching the state `valid`, and the rest which are all the properties and state. Party of watchers 🥳 
This is a no-go.

I would create the following.
```javascript
computed: mapState('globalErrorsModule', ['valid'])
```
then in the watchers piece 
```javascript
watch: {
  "valid": {
      this.isInvalid = !valid;
    }
}
```

josechirivella (changes requested):

src/components/renderer/form-button.vue
caleeli: Thanks! fixed

josechirivella (approved):

caleeli:
The failing test was using a bad selector `#__BVID__95___BV_modal_footer_` changed to `[data-cy="modal-remove"]`.
I checked that the generated test screen has the same html structure to the previous of the change, but with different `__BVID__`.

image

src/components/renderer/form-button.vue
```diff
@@ -9,40 +9,25 @@
 <script>
 import Mustache from 'mustache';
 import { getValidPath } from '@/mixins';
-import { isProxy } from 'is-proxy';
```
eiresendez, line 12: @caleeli @josechirivella hey guys! Please consider removing this `"is-proxy": "^1.0.6",` dependency, since it was only used in this file. 👍

caleeli: @eiresendez thanks, dependency was removed.

eiresendez (approved):
Awesome!

agustinbusso (commented):
- [x]  Screen with visibility rule and validations. Test writing and submitting fast
- [x]  Screen with visibility rule and validations. Test validation is working on submit
- [x]  Screen with visibility rule and validations. Test validation debounced is correctly working

https://user-images.githubusercontent.com/90727999/191038129-e1e89a84-aeba-4239-b10d-3941cc2a66f2.mov

In screens with nested screens refresh validations continue triggering multiple times, but debounced. There is a way to improve that?

https://user-images.githubusercontent.com/90727999/191043513-804609a4-31d2-44a8-bb42-da1dc799b548.mov

caleeli:
@agustinbusso that's weird, the event `refresh-validation-rules` was completely removed from code, so it should not be triggered any more from any component.
Now when multiple nested-screens triggers requires the validation refresh, only the last one is served (by debounce)

agustinbusso:
Refresh validation are not triggering anymore.

Maybe we can get an unexpected behavior when a user write fast, the validation maybe will not work correctly.
Also notice when you fill the three inputs, and remove the text in some of them, and quickly submit the form, an empty error message shows.

https://user-images.githubusercontent.com/90727999/191056795-a0aef739-969a-4335-a2a0-d14e0dcc726c.mov

Also I can notice a good performance improvement in heavy screens with nested screens:

https://user-images.githubusercontent.com/90727999/191057899-5844625d-559b-41d7-9c62-ba7452a8dd68.mov

agustinbusso (approved):
**Validation refreshing correctly when submitting the form**

https://user-images.githubusercontent.com/90727999/191606765-2836e290-61d7-4b72-8fe8-1211bdba1a9b.mov

PR #1257, opened 2022-09-23
Four 6732/change vuejsonpretty

caleeli:
## Issue & Reproduction Steps
Preview Data in Screen Builder is refreshing many times when a input changes.

## Solution
Delay and debounce the refresh of data preview

## How to Test
- Create a form with some inputs and other controls
- Go to Preview
- Focus on a input text
- Type
- Note in the Preview Data panel VueJosnPretty was changed by Monaco
- Note the Preview Data is devounced (refresh after user complete to type)

https://user-images.githubusercontent.com/8028650/191998125-ad117d9a-bb43-4801-953c-15a6a32266a3.mp4

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-6732

## Code Review Checklist

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -78,6 +78,7 @@
               :key="rendererKey"
               v-model="previewData"
               @submit="previewSubmit"
+              @update="updateDataPreview"
               :mode="mode"
               :config="preview.config"
               :computed="preview.computed"
@@ -101,7 +102,14 @@
                 </b-button>
 
                 <b-collapse v-model="showDataInput" id="showDataInput">
-                  <monaco-editor :options="monacoOptions" class="data-collapse" v-model="previewInput" language="json" data-cy="preview-data-input"/>
+                  <monaco-editor
+                    v-model="previewInput"
+                    :options="monacoOptions"
+                    class="data-collapse"
+                    language="json"
+                    data-cy="preview-data-input"
+                    @change="updateDataInput"
+                  />
                 </b-collapse>
 
                 <b-button variant="outline"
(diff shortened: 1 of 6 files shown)


src/App.vue
```diff
@@ -412,6 +430,13 @@ export default {
     this.loadFromLocalStorage();
   },
   methods: {
+    updateDataPreview() {
+      this.previewDataStringify = JSON.stringify(this.previewData, null, 2);
+    },
```
josechirivella, line 435: in order to remove the created on line 409 and to make this method to look nice and have the debounce functionality. I would write it like...
```suggestion
    updateDataPreview: debounce(function () {
      this.previewDataStringify = JSON.stringify(this.previewData, null, 2);
    }, 1000),
```

src/App.vue
```diff
@@ -315,6 +324,12 @@ export default {
     WatchersPopup,
   },
   watch: {
+    previewData: {
+      deep: true,
```
josechirivella, line 328: do we really need to watch every single property?

src/App.vue
josechirivella: if the following suggestion is approved. You can remove the created on line 409:411

josechirivella (changes requested):

src/App.vue
caleeli: We need to react to any change and refresh the preview data, but I will review a better way to do it.

src/App.vue
caleeli: Fixed

src/App.vue
caleeli: Removed the watchers, and using events to refresh the preview data

src/App.vue
```diff
@@ -412,6 +416,22 @@ export default {
     this.loadFromLocalStorage();
   },
   methods: {
+    // eslint-disable-next-line func-names
```
josechirivella, line 419: if you were getting an eslint-error, then leave this line as it is. But, if not, you were getting an eslint warning which you can remove this line. Same goes to line 427

josechirivella (approved):
Only had one comment/suggestion but everything else looks good

src/App.vue
caleeli: I agree with the previous observation, it is much better (or the right way) to "debounce" this method in its definition rather that changing it during the component creation. About the `eslint-error` maybe that could be changed in a different ticket.

PR #1260, opened 2022-10-12
Four 6749

rodriquelca:
## Issue & Reproduction Steps
A select list into a loop make services call as the loop **Default Loop Count** says causing repeated calls.

## Solution
We added an axios extension to handle the services and cache 
The code included in this PR modify this problems in components:
- SelectList

## How to Test
- Test with a single screens, with select list inside a loop
- Configure a Data source for the select list 
- Go to preview
- Open the developer tool -> network
- Review if the service is cached 

## Related Tickets & Packages
- Required by: 
- https://processmaker.atlassian.net/browse/FOUR-6749

## Code Review Checklist


src/DataProvider.js
```diff
@@ -136,19 +161,45 @@ export default {
...
+   * @param {number|null} dataSourceId
+   * @param {object} params
+   * @returns {object}
+   */
+  getDataSource(dataSourceId, params) {
+    let url = `/requests/data_sources/${dataSourceId}/resources/${params.config.endpoint}/data`;
+    url += this.authQueryString();
+    return this.get(url, {
+      useCache: window.ProcessMaker.screen.cacheEnabled,
+      params: params.config.outboundConfig
```
caleeli, line 178: Missing `params.data`.
The endpoint `/api/1.0/requests/data_sources/{data_source_id}` could receive the following parameters:

In GET: `?queryString=some_string`
In POST: ` config: { object } `
In POST: ` data: { object } `

Since the `config` and `data` could be complex objects, consider to encode with JSON the parameters `config` and `data`. Something like:

`?config={"params":{"foo":"bar"}}&data={"foo":"bar"}`

josechirivella (approved):

caleeli (approved):

PR #1282, opened 2022-11-11
FIX: Unable to access _parent parameter on default value JS in line input and Nested Screen FOUR-7029

caleeli:
## Issue & Reproduction Steps
1. Create a Screen Main
2. Add line input as parent value
3. Add loop control
4. Add Nested Screen 
5. Create a Screen (Nested Screen)
6. Add line input
7. Configure default value of line input as: `return this._parent.parentInput;`
8. Press preview button
9. Write some text on line input parent

## Current Behavior 

_parent is not working in default values  

## Expected Behavior 

_parent should work  in default value inside a nested screen as PM 4.2.35. As show in the video  

## Solution
- Fix the proxy to access Data

## How to Test
Test the steps above

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-7029

## Code Review Checklist


src/mixins/DataReference.js
```diff
@@ -0,0 +1,85 @@
+/**
+ * Gets the screen parent or null if don't have
+ * @returns {object|null}
+ */
+function findScreenOwner(control) {
```
josechirivella, line 5: So, I feel like this file is the same one I've been seeing across the PRs from this week. Is there any way we can have one single file to serve multiple scenarios? To avoid duplication?

josechirivella (changes requested):

src/mixins/DataReference.js
josechirivella: In agreement, @caleeli is going to solve this issue in another ticket.

rodriquelca (approved):

ryancooley (approved):

PR #1291, opened 2022-11-21
Signature not working in an image control inside a loop

marcoAntonioNina:
## Issue & Reproduction Steps
1. import the screen in the ticket
2. perform a signature with the signature component.
3. checking the loop and form image component, the image should be the same.
Expected behavior: 
variable _parent should be working with FormImage component
Actual behavior: 
variable _parent does not return the information
## Solution
- Add method get screen reference

## How to Test
use variable _parent for load data in FormImage Component

## Related Tickets & Packages
[FOUR-7108](https://processmaker.atlassian.net/browse/FOUR-7108)

## Code Review Checklist


sanjacornelius (approved):

src/components/renderer/form-image.vue
```diff
@@ -18,7 +18,9 @@ export default {
   },
   computed: {
     imageUrl() {
-      return get(this.validationData, this.variableName || this.imageName);
+      if (this.mode === 'preview' ) {
```
caleeli, line 21: This will only work in preview mode. It will always return undefined when you run a process inside a Request Task

caleeli (changes requested):
Please fix the case when the screen is used outside the editor. (Runtime)

caleeli (approved):

ryancooley (approved):

PR #1319, opened 2023-01-20
FOUR-7000: Improve Submit button component

rodriquelca:
## Issue & Reproduction Steps

Expected behavior: 
submit must work with improvements

## Solution
- some improvements was added
- test was included

## How to Test
Test the steps above

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-7000

## Code Review Checklist

### DIFF
--- a/src/components/renderer/form-button.vue
+++ b/src/components/renderer/form-button.vue
@@ -1,49 +1,100 @@
 <template>
-  <div class="form-group"  style="overflow-x: hidden">
-    <button v-b-tooltip="options" @click="click" :class="classList" :name="name" :aria-label="$attrs['aria-label']" :tabindex="$attrs['tabindex']">
+  <div class="form-group" style="overflow-x: hidden">
+    <button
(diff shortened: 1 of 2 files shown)


src/components/renderer/form-button.vue
```diff
@@ -1,49 +1,100 @@
...
+      type: String,
+      default: null
+    },
+    fieldValue: {
+      type: String,
+      default: "1"
+    },
+    value: {
+      type: String,
+      default: "1"
```
josechirivella, line 49: I don't think we want the default value and fieldValue to be `1` right?

josechirivella (changes requested):
and the test are failling

src/components/renderer/form-button.vue
rodriquelca: right props types and defaults was removed because has causing some compatibilitie errors

caleeli (approved):

PR #1332, opened 2023-02-09
FOUR-6851: Improve/refactor Image control

rodriquelca:
## Issue & Reproduction Steps

Expected behavior: 
handle repeated images as a blob files for performance issues
Actual behavior: 
images has performance issues into a loop
## Solution
- image 64 bit was transformed to a blob files for loops control

## How to Test
Test the steps above

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-6851

## Code Review Checklist


src/mixins/extensions/LoadFieldComponents.js
```diff
@@ -27,6 +27,33 @@ export default {
...
       });
     },
+    createObjectURL(base64image) {
+      if (base64image) {
+        const binaryData = atob(base64image.split(",")[1]);
+        const binaryArray = new Uint8Array(binaryData.length);
+        for (let i = 0; i < binaryData.length; i++) {
+          binaryArray[i] = binaryData.charCodeAt(i);
+        }
+        const blob = new Blob([binaryArray], { type: "image/jpeg" });
```
josechirivella, line 37: Hardcoding the extension file can be a double edge sword. Maybe we need to get the file extension using regex or whatever and use that value here. What do you think @caleeli

josechirivella (changes requested):

src/mixins/extensions/LoadFieldComponents.js
caleeli: Agree!

caleeli (changes requested):
Please review the comments

PR #1334, opened 2023-02-10
Upgraded Cypress to v9

josechirivella:
## Issue & Reproduction Steps

Expected behavior: 
The ability to run all of our tests in Cypress v9

Actual behavior: 
Currently, our Cypress version is v3 which is pretty old

## Solution
- Upgrading Cypress to v9
- Some Cypress utilities were deprecated, I had to upgrade those manually like `cy.intercept` and some `should` assertions
- I had to move the cypress folder from `tests/` to his own folder in the root folder since that's the new structure recommended from Cypress

## How to Test
Test the steps above
- Run `npm run open-cypress` after spinning up the `npm run serve`

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-6510

## Code Review Checklist


cypress/integration/Task.spec.js
```diff
@@ -212,11 +207,19 @@ describe('Task component', () => {
     });
     cy.wait(2000);
     cy.get('.form-group').find('button').click();
-    cy.route('PUT', 'http://localhost:8080/api/1.0/tasks/1').then(function() {
+    /* TODO This call is failing with a 404 . Debug on the why. I could trace that the line 221, user_id was being sent
```
caleeli, line 210: This is failing because the following Route is not being intercepted, also note that the URL is malformed (`user_id=null`, `process_request_id=null`):
```
http://localhost:8080/api/1.0/tasks?user_id=null&status=ACTIVE&process_request_id=null&include_sub_tasks=1
```
Prior to this PR, it also returned 404, but Cypress did not end the test in a failed state.
image

The following ticket was created to fix this issue: https://processmaker.atlassian.net/browse/FOUR-7412

caleeli (approved):

PR #1377, opened 2023-05-23
FOUR-8060 Interstitial Reroute

gproly:
## Issue & Reproduction Steps
Create a process with: start->task->script->task->end
In the script, put throw new \Exception('Custom error');
The first task should have the interstitial configured.

Expected behavior: 
Now, only a simple error message must be displayed.

Actual behavior: 
We redirect the user to the error detail page

## Solution
Change the redirection behaviour

### DIFF
--- a/src/components/SimpleErrorMessage.vue
+++ b/src/components/SimpleErrorMessage.vue
@@ -1,0 +1,43 @@
+<template>
+    <div class="simple-error-message-main">
+        <p>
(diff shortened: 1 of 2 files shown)


nolanpro (approved):

pmoskrojas (approved):
The following code to the functionality of the ticket FOUR-8060, meets desired.

PR #1379, opened 2023-05-24
FOUR-8509: Required fields in nested screens are lost 

danloa:
## Issue & Reproduction Steps

- Create a Form Screen with two pages at least
- Include on all pages the required fields
- Create other Form Screen
- Add a nested screen and assign the previous form.
- Add a Submit button
- Add a second nested screen with no screen assigned
- Go to Preview mode. All required fields are required on the first page, but on other pages lost all required fields.
- Create a process
- Add a Start Event, form task, and End Event.
- Connect one by one.
- Save all changes
- Start a new request for the process
- The task started and all required are required on the first page, but on other pages lost all required fields and it is possible to submit the task.

Expected behavior: 
In a nested screen that have more than 1 page, validation of all pages should be used.

Actual behavior: 
Validations of the only the first page are used.

## Solution
When  nested screen validations were added, the main screen definition was passed to the ValidationFactory. This was fixed, now the nested screen is passed, so that all the nested screen validations are added.

## Related Tickets & Packages
- [https://processmaker.atlassian.net/browse/FOUR-8509](https://processmaker.atlassian.net/browse/FOUR-8509)

## Code Review Checklist

### DIFF
--- a/src/ValidationsFactory.js
+++ b/src/ValidationsFactory.js
@@ -85,8 +85,13 @@ class FormNestedScreenValidations extends Validations {
     }
     const definition = await this.loadScreen(this.element.config.screen);
     let parentVisibilityRule = this.parentVisibilityRule ? this.parentVisibilityRule : this.element.config.conditionalHide;
-    if (definition && definition[0] && definition[0].items) {
-      await ValidationsFactory(definition[0].items, { screen: this.screen, data: this.data, parentVisibilityRule }).addValidations(validations);
+
+    if (Array.isArray(definition)) {
+      for(var i = 0; i < definition.length; i++) {
+        if (definition[i].items) {
+          await ValidationsFactory(definition[i].items, { screen: this.screen, data: this.data, parentVisibilityRule }).addValidations(validations);
+        }
+      }
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/ValidationsFactory.js
```diff
@@ -83,13 +83,30 @@ class FormNestedScreenValidations extends Validations {
     if (!this.isVisible()) {
       return;
     }
-    const definition = await this.loadScreen(this.element.config.screen);
+    const nestedScreen = await this.loadNestedScreen(this.element.config.screen);
+    //const definition = await this.loadScreen(this.element.config.screen);
```
caleeli, line 87: Remove commented code

src/ValidationsFactory.js
```diff
@@ -83,13 +83,29 @@ class FormNestedScreenValidations extends Validations {
     if (!this.isVisible()) {
       return;
     }
-    const definition = await this.loadScreen(this.element.config.screen);
+    const nestedScreen = await this.loadNestedScreen(this.element.config.screen);
+    const definition = nestedScreen.config;
```
caleeli, line 87: Add validation here since `nestedScreen` could be empty

caleeli (approved):
The change is good, but consider to move the variable `pagesValidated` into a class, because this ValidationFactory can be called asynchrnosly multiple times because of the reactivity.

PR #1385, opened 2023-06-16
FOUR-7964: Screen-size Visibility Rules

danloa:
Implements:
[https://processmaker.atlassian.net/browse/FOUR-7964](https://processmaker.atlassian.net/browse/FOUR-7964)

### DIFF
--- a/src/components/inspector/device-visibility.vue
+++ b/src/components/inspector/device-visibility.vue
@@ -26,8 +26,13 @@
 
 <script>
 export default {
-  components: {},
-  props: ["label", "value", "helper", "options"],
+  // props: ["label", "value", "helper", "options"],
+  props: {
+    label: String,
+    value: String,
+    helper: String,
+    options: Object,
+  },
   data() {
     return {
       settings: {
--- a/src/components/inspector/device-visibility.vue
+++ b/src/components/inspector/device-visibility.vue
@@ -4,7 +4,7 @@
       <label>{{ label }}</label>
 
       <b-form-checkbox
-        :checked="value.showForDesktop"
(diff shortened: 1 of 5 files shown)


src/components/inspector/device-visibility.vue
```diff
@@ -0,0 +1,64 @@
...
+      >
+        {{ $t("Show for Mobile") }}
+      </b-form-checkbox>
+    </div>
+  </div>
+</template>
+
+<script>
+export default {
+  components: {},
```
josechirivella, line 29: remove if not using

src/components/inspector/device-visibility.vue
```diff
@@ -0,0 +1,64 @@
...
+  props: ["label", "value", "helper", "options"],
```
josechirivella, line 30: can you add the prop types?

src/components/inspector/device-visibility.vue
```diff
@@ -0,0 +1,64 @@
...
+  data() {
+    return {
+      settings: {
+        showForDesktop: true,
+        showForMobile: true
+      }
+    };
+  },
+  watch: {
+    settings: {
```
josechirivella, line 40: can a computed property be possible here? Trying to avoid watcher

josechirivella (changes requested):

josechirivella (approved):

src/components/inspector/device-visibility.vue
josechirivella: oh, you can add a @change property to it. If the v-model is ever changed, that would trigger the response

josechirivella (approved):

PR #1463, opened 2023-10-13
FOUR-10340: Designer: The list needs to updated according to the filter applied

CarliPinell:
## Solution
Functionalities in all filters in List Table control were applied 

## Related Tickets & Packages
- FOUR-10340 processmaker

## Code Review Checklist
ci:next


src/components/renderer/form-new-request.vue
```diff
@@ -107,6 +108,15 @@ export default {
           this.processes[category.name].push(process);
         }
       }
+    },
+    fetchData(selectedOption) {
```
fagubla, line 112: Start new Requests is not necessary the filters

fagubla (changes requested):

src/components/renderer/form-new-request.vue
CarliPinell: Solved.

fagubla (approved):

PR #1490, opened 2023-12-01
add openAcordeon method and update tests

karlOS123testing:
fix some test cases in order to open the acordeon


pmoskrojas (approved):
All the changes related to the new Accordion component in the Control Menu, perfectly comply with what is required on the Cypress test. Therefore, the work is excellent and approved.

PR #1492, opened 2023-12-06
qa_fix_testCases

karlOS123testing:


pmoskrojas (approved):
The changes made are relevant to the solution of the test of the new component and behavior, found in Control Menu. Therefore, they are approved.

PR #1509, opened 2024-01-08
Fix_tests_qa

karlOS123testing:
`div` tag with `overflow-auto` class affects in the 'should' commands inside the `openAcordeon` method


karlOS123testing:
I closed this PR because it was to a wrong branch.

PR #1552, opened 2024-02-28
FOUR-14223 the correct tabs are not closed when after deleting the first tab

devmiguelangel:
## Issue & Reproduction Steps

Expected behavior: 
The tab where x is clicked should be closed (in this case the second tab)

Actual behavior: 
The last tabs are closed consecutively but not the second tab

## Solution
send the page number instead of the index to the `closeTab` method

## How to Test
1. Import the attached screen
2. Open screen
3. Display 6 more pages like tabs
4. Go to the last tab
5. Close first tab
6. Try closing second tab(Click on x on the second tab 5 times)

## Related Tickets & Packages
[FOUR-14223](https://processmaker.atlassian.net/browse/FOUR-14223)

## Code Review Checklist

ci:next

[FOUR-14223]: https://processmaker.atlassian.net/browse/FOUR-14223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/vue-form-builder.vue
+++ b/src/components/vue-form-builder.vue
@@ -382,6 +382,7 @@
         @item-edit="() => {}"
         @item-delete="confirmDelete"
         @add-page="$bvModal.show('addPageModal')"
+        @ordered="orderPages"
       />
     </b-modal>
 
@@ -1352,7 +1353,9 @@ export default {
       this.updateState();
       this.inspect(clone);
     },
-
+    orderPages(items) {
+      this.config = [...items];
+    },
   }
 };
 </script>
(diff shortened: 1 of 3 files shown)


src/components/vue-form-builder.vue
```diff
@@ -1352,7 +1353,9 @@ export default {
       this.updateState();
       this.inspect(clone);
     },
-
+    orderPages(items) {
+      this.config = [...items];
```
caleeli, line 1357: Change the order of the pages directly to the screen config is a risk, need to recalculate the references in record lists and navigation buttons, I think the order property for each page is enough.

Add the `this.updateState();` to save the changes in the state (it also needed for undo/redo)

PR #1662, opened 2024-08-01
FOUR-17488: Back redirection on element destination with external URL is not correct

rodriquelca:
## Issue & Reproduction Steps

Expected behavior: 
When clicking back in the browser, it should not redirect to the process editing and then return to the same URL....

It should be defined what would be the appropriate behavior for this use case.

Actual behavior: 
As we see, when we click back in the browser it redirects us to the process editing and a few seconds later it returns to the external URL

## Solution
- add sessionUrlActionBlocker to prevent the load form action

https://github.com/user-attachments/assets/dd977a7f-2a5b-474e-8e86-86474b943f87

## How to Test
Test the steps above

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-17488

## Code Review Checklist

ci:next
ci:vue-form-elements:next


src/components/task.vue
```diff
@@ -797,6 +798,30 @@ export default {
...
+     * Checks for the presence of a URL action blocker in sessionStorage and handles it.
+     *
+     * This method retrieves the 'sessionUrlActionBlocker' value from sessionStorage,
+     * and if present, removes it and emits a 'closed' event with the task id and 
+     * the source of the redirection. It returns false if the blocker was handled, 
+     * and true otherwise.
+     *
+     * @returns {boolean} Returns false if the 'sessionUrlActionBlocker' was found and handled, true otherwise.
+     */
+    hasUrlActionBloker() {
```
caleeli, line 811: Typo missing `c`
```suggestion
    hasUrlActionBlocker() {
```

src/components/task.vue
```diff
@@ -807,7 +832,9 @@ export default {
     this.nodeId = this.initialNodeId;
     this.requestData = this.value;
     this.loopContext = this.initialLoopContext;
-    this.loadTask();
+    if (!this.hasUrlActionBloker()) {
```
caleeli, line 835: Typo missing `c`

src/components/task.vue
```diff
@@ -455,6 +455,7 @@ export default {
 
       const elementDestinationUrl = elementDestination.value;
       if (elementDestinationUrl) {
+        sessionStorage.setItem('sessionUrlActionBlocker', document.referrer);
```
caleeli, line 458: Similar as the code below consider to add some comment to document the meaning of `sessionUrlActionBlocker`
```
// Save the referring URL to sessionStorage for future verification
```

devmiguelangel (approved):

PR #1686, opened 2024-09-03
FOUR-18067 STORY Add Encrypted Toggle in the Input Field

julceslauhub:
https://processmaker.atlassian.net/browse/FOUR-18067

ci:next

### DIFF
--- a/src/components/inspector/encrypted-config.vue
+++ b/src/components/inspector/encrypted-config.vue
@@ -1,0 +1,56 @@
+<template>
+  
+    <div>
+      <div class="form-group">
+        <b-form-checkbox
+          v-model="settings.encrypted"
+          @change="emitChanges"
+          switch
+        >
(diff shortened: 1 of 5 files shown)


src/components/inspector/encrypted-config.vue
```diff
@@ -0,0 +1,56 @@
...
+    value: {
+      type: Object,
+      default: () => {
+        return { encrypted: false };
+      }
+    },
+    helper: String,
+  },
+  watch: {
+    value: {
```
caleeli, line 30: is it possible to do not use watcher? it cause the reactivity to be hard to track

julceslauhub: this snippet of code is similar to all another fields attributes used in the srceen builder

caleeli (approved):

rodriquelca (approved):

PR #1765, opened 2024-10-28
FOUR-19778 Encrypted field does not have read-only mode

julceslauhub:
## Issue & Reproduction Steps

Expected behavior: 
"readonly" attribute should be considered when a value is revealed/plain

Actual behavior: 
"readonly" attribute is omitted

## Solution
Keep a copy of the original values in order to restore the correct "readonly" mode

## Related Tickets & Packages
https://processmaker.atlassian.net/browse/FOUR-19778

## Code Review Checklist

### DIFF
--- a/src/components/renderer/form-masked-input.vue
+++ b/src/components/renderer/form-masked-input.vue
@@ -41,7 +41,7 @@
       <input
         v-else
         v-model="localValue"
-        v-bind="componentConfig"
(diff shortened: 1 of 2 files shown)


caleeli (changes requested):
@julceslauhub please include the test for the Encrypted Field + Read-only scenario

caleeli (approved):

PR #1789, opened 2025-02-04
FOUR-21660: The form screen is shown as conversational when we start a new case

eiresendez:
## Issue & Reproduction Steps

Expected behavior: 
- If we have two tasks and each one has a different type of screen, the screens should not change type in execution mode.

Actual behavior: 
- The conversational screen that is in the first task is completed and the controls of the next form screen are being displayed as if they were part of the conversational screen.

## Solution
- The issue was caused by the logic flow in handleRedirectToTask(), where loadTask() was being used instead of a hard redirect.
- The fix was to force a navigation to the correct task screen using window.location.href

## How to Test
- Follow the reproduction steps from the JIRA ticket.

## Related Tickets & Packages
- [FOUR-21660](https://processmaker.atlassian.net/browse/FOUR-21660)

## Code Review Checklist

[FOUR-21660]: https://processmaker.atlassian.net/browse/FOUR-21660?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/task.vue
+++ b/src/components/task.vue
@@ -890,6 +890,8 @@ export default {
         if (this.renderComponent === "ConversationalForm") {
           window.location.href = `/tasks/${this.taskId}/edit`;
         }
+
+        this.reload();
       }
     },
 
--- a/src/components/task.vue
+++ b/src/components/task.vue
@@ -878,16 +878,18 @@ export default {
         this.loadingTask = true;
         // Check if interstitial tasks are allowed for this task.
         if (this.task && !(this.task.allow_interstitial || this.isSameUser(this.task, data))) {
-           // The getDestinationUrl() function is called asynchronously to retrieve the URL
+          // The getDestinationUrl() function is called asynchronously to retrieve the URL
           window.location.href = await this.getDestinationUrl();
           return;
         }
         this.nodeId = data.params[0].nodeId;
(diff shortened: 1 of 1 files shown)


src/components/task.vue
```diff
@@ -884,7 +884,10 @@ export default {
         }
         this.nodeId = data.params[0].nodeId;
         this.taskId = data.params[0].tokenId;
-        this.reload();
+
+        // Force a redirect to ensure the correct task is loaded immediately.
+        // This prevents async reloads that may cause inconsistencies.
+        window.location.href = `/tasks/${this.taskId}/edit`;
```
caleeli, line 890: This change could affect Interstitial and Webentry, this uses cases must be verified.

caleeli: Maybe and IF statement could be added to limit the scope of the change

boliviacoca:
Added HOLD because of @caleeli comment, asked PO to have it moved to a further release.

src/components/task.vue
eiresendez: Ty David, done!

marcoAntonioNina (approved):

PR #1906, opened 2026-03-27
FOUR-30113: Interstitial screen keeps loading with scripts between form tasks with self service and group assignment

eiresendez:
## Issue & Reproduction Steps

Expected behavior:
When the interstitial cannot find a next active task for the current user, but there is still an active task in the request for another user/group, it should stop waiting and take the user back to the request instead of staying in the interstitial forever.

Actual behavior:
With a flow like `Start Event -> FT-A (self service / Requesters) -> Script Task -> FT-B`, the start-event interstitial keeps showing "We're getting the next task for you..." forever for `admin`, even though the request already has an active self-service task that exists in the background and is not claimable by the current user.

## Solution
- Add a second active-task lookup in the interstitial flow when the user-filtered lookup returns no tasks.
- If another active task exists in the same request for a different user/group, redirect to `/requests/{requestId}` instead of waiting forever.
- Keep the existing behavior when a task is available for the current user, and keep waiting when no active task exists yet.
- Add e2e coverage for the new fallback and for the legitimate wait case.
- Add regression coverage showing that claimable self-service tasks still resolve through the existing user-filtered lookup and `redirectToTask` + `userCanClaim` paths.

Scope note:
This PR is intentionally scoped to the reproduced interstitial fallback issue above. It does not attempt to infer claimability from the second unfiltered lookup. If FOUR-30113 needs that broader behavior, it should be handled in a follow-up with explicit backend/API support.

## How to Test
Test the steps above

Manual validation used for this proposed fix:
1. Use a process like `Start Event -> FT-A -> Script Task -> FT-B`.
2. Configure `FT-A` as self-service for a group the current user does not belong to.
3. Start a new case as that current user.
4. Confirm the app redirects to the request instead of staying forever in the interstitial.
5. Confirm the request shows the active self-service task in the task list.

## Related Tickets & Packages
- [FOUR-30113](https://processmaker.atlassian.net/browse/FOUR-30113)

## Code Review Checklist

ci:deploy

[FOUR-30113]: https://processmaker.atlassian.net/browse/FOUR-30113?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/task.vue
+++ b/src/components/task.vue
@@ -526,40 +526,64 @@ export default {
       // If the task has not an origin source it should re redirected top the tasks List as default.
       return document.referrer || '/tasks';
     },
-    loadNextAssignedTask(requestId = null) {
+    getActiveTasksUrl(requestId, includeUserFilter = true) {
+      const timestamp = !window.Cypress ? `&t=${Date.now()}` : "";
+      const userFilter = includeUserFilter ? `user_id=${this.userId}&` : "";
+
+      return `?${userFilter}status=ACTIVE&process_request_id=${requestId}&include_sub_tasks=1${timestamp}`;
+    },
+    isWaitingOnInterstitial() {
+      return Boolean(this.task?.allow_interstitial && _.get(this.screen, '_interstitial', false));
+    },
+    async redirectToRequestIfAnotherActiveTaskExists(requestId) {
+      if (!this.isWaitingOnInterstitial()) {
+        return false;
+      }
+
+      const response = await this.$dataProvider.getTasks(this.getActiveTasksUrl(requestId, false));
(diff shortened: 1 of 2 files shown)


Kookster310:
QA server K8S was successfully deployed https://ci-40377ccf4c.engk8s.processmaker.net

eiresendez:
https://github.com/user-attachments/assets/f7baddc6-89c1-4d18-b86a-7ee9a20c6caa

eiresendez:
Closing PR as this fix is no longer necessary. Final decision aligned with the team.
cc. @caleeli @gproly @sanjacornelius

PR #1923, opened 2026-07-27
Record list does not retrieve the selected value/data in the next task.

CarliPinell:
## Solution
- When a record is selected from the Record List, the value persists when moving to the next task.

## How to Test
1. Create/import process attachaed in ticket
2. Create a simple screen
3. Add record list and add a collection to the configuration
4. Create a case, select an option in the record list, complete task 1

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-27264

## Code Review Checklist

ci:deploy

### DIFF
--- a/src/components/renderer/form-record-list.vue
+++ b/src/components/renderer/form-record-list.vue
@@ -646,11 +646,7 @@ export default {
     onRadioChange(selectedItem, index) {
       const globalIndex = (this.currentPage - 1) * this.perPage + index;
       // Prefer dataSelectionOptions; fall back to singleField for legacy configs.
-      const isSingleField =
-        this.source?.dataSelectionOptions === "single-field" ||
-        (this.source?.dataSelectionOptions == null && !!this.source?.singleField);
-
-      if (isSingleField && this.source?.singleField) {
+      if (this.isSingleFieldSelectionMode() && this.source?.singleField) {
         this.componentOutput(this.getSingleFieldValue(selectedItem));
       } else {
         this.componentOutput({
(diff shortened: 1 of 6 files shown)


src/components/renderer/form-record-list.vue
```diff
@@ -472,14 +473,42 @@ export default {
...
+      }
+      if (
+        !["single-field", "single-record"].includes(
+          this.source?.dataSelectionOptions
+        )
+      ) {
+        return;
+      }
+      const pageIndex = Array.isArray(this.tableData?.data)
+        ? this.tableData.data.indexOf(newVal)
```
eiresendez: 🔴 [src/components/renderer/form-record-list.vue:493](/Users/eresendez/projects/packages/processmaker/screen-builder/src/components/renderer/form-record-list.vue:493) The new `selectedRow` watcher duplicates the existing `@change` path and passes a global array index into `onRadioChange()`, which expects a page-relative index. On page 2 with five rows per page, selecting global row 5 temporarily persists `selectedRowIndex: 10` instead of `5`. BootstrapVue emits both the model event and `change`, so one click can update `vdata` twice—first with incorrect metadata—potentially triggering ProcessMaker screen watchers or dependent PMQL requests twice.
  - Suggested fix: use one selection event path. If the watcher remains, pass a page-relative index or persist the already-global index without adding the page offset again.

tests/unit/FormRecordListCollectionSelection.spec.js
```diff
@@ -0,0 +1,299 @@
+/**
```
eiresendez, line 1: 🟡 Nice-to-have (Optional improvement)
- The tests duplicate simplified implementations instead of importing or mounting FormRecordList. Consequently, they cannot detect the real watcher/@change interaction or failures caused by production code drifting from the copied helpers.

eiresendez (changes requested):
@CarliPinell the fix addresses the original issue, and I confirmed it no longer reproduces in manual testing. 

However, this PR introduces a new pagination issue through the added `selectedRow` watcher: the existing `@change` path already provides a page-relative index, while the watcher passes a global index into the same handler, causing an incorrect `selectedRowIndex` and duplicate updates on page 2 and beyond. This behavior was not present before this change, could you please review it?

src/components/renderer/form-record-list.vue
CarliPinell: Suggested fix done.

tests/unit/FormRecordListCollectionSelection.spec.js
CarliPinell: Done

eiresendez (approved):
The new helper extraction is clean and keeps the component’s selection contract explicit. [form-record-list.vue:626](https://github.com/ProcessMaker/screen-builder/blob/783df9dad483cec6c7ca374b9afa4ca38a4b9110/src/components/renderer/form-record-list.vue#L626) now receives a page-relative index and delegates the single conversion to production-tested code 👍 👍

nolanpro:
QA server K8S was successfully deployed https://ci-0e0ae22cfc.engk8s.processmaker.net

eiresendez (approved):
The merge resolution correctly contracts: it keeps the non-mutating record mapping and legacy data.* compatibility. The previously approved page-relative index behavior remains unchanged. 👍 👍

PR #1930, opened 2026-08-26
FOUR-32834: Preserve Collection Record Control mode

eiresendez:
## Issue & Reproduction Steps

When two Collection Record Controls use the same dynamic Record ID but different modes, resolving the Record ID causes both controls to display the Edit Screen.

To reproduce, configure two controls against the same collection using `{{option}}` as the Record ID, set one to Edit and the other to View, open Preview, and enter a valid record ID.

Expected behavior: Each control preserves its configured mode and loads the corresponding Update or Read Screen.

Actual behavior: Both controls load the Update Screen in Edit mode.

## Solution

- Resolve the effective mode from `collectionmode.modeId` whenever the dynamic Record ID changes.
- Reuse the same mode resolution during initial mounting and subsequent record changes.
- Preserve Edit as the fallback for legacy configurations without a collection mode.
- Add regression coverage for controls with Edit and View modes sharing a dynamic Record ID.

## How to Test

- Run `npx jest tests/unit/FormCollectionRecordControl.spec.js --runInBand`.
- Run ESLint against the new resolver and regression test.
- Run `npm run build-bundle`.
- Copy the generated `dist/` contents into Core and run `npm run dev`.
- In Screen 17 Preview, enter `1` in `option` and verify that the first control remains editable while the second displays its Read Screen.
- The focused tests pass. The full Jest run reports 94 passing tests, while 7 existing suites fail before execution because of the repository's Vue/Jest transformation configuration.

## Related Tickets & Packages

- https://processmaker.atlassian.net/browse/FOUR-32834
- Package: `@processmaker/screen-builder`

ci:deploy

### DIFF
--- a/src/components/renderer/collection-record-mode.js
+++ b/src/components/renderer/collection-record-mode.js
@@ -1,0 +1,6 @@
+export default function resolveCollectionMode(
+  collectionMode,
+  defaultMode = "Edit"
+) {
+  return collectionMode?.modeId ?? defaultMode;
+}
--- a/src/components/renderer/form-collection-record-control.vue
+++ b/src/components/renderer/form-collection-record-control.vue
@@ -17,6 +17,7 @@
 import _ from "lodash";
 import VueFormRenderer from "../vue-form-renderer.vue";
 import CollectionRecordsList from "../inspector/collection-records-list.vue";
+import resolveCollectionMode from "./collection-record-mode";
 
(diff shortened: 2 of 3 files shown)


nolanpro:
QA server K8S was successfully deployed https://ci-178ddcc9bb.engk8s.processmaker.net

mcraeteisha (approved):

PR #1931, opened 2026-08-27
FOUR-32798: Preserve Record List collection configuration

eiresendez:
## Issue & Reproduction Steps

A Collection-backed Record List can lose its Data Selection, selected single field, and configured columns in two related scenarios:

1. Import or link a Process whose Collection ID is remapped while the Record List still contains cached records from the source Collection, then open the imported Screen.
2. Configure columns for a Collection with zero records, switch from Design to Preview, and return to Design.

Expected behavior: The Record List preserves its persisted Data Selection, legacy `data.*` single-field value, and configured columns while the same Collection is hydrated or refreshed.

Actual behavior: Screen Builder interprets stale or empty record data as a user-initiated Collection change and clears valid configuration.

## Solution

- Emit the Collection change signal only from an explicit user selection instead of inferring it from loaded records.
- Preserve Data Selection and `singleField` during hydration, clearing `singleField` only when the user manually changes Data Selection.
- Normalize legacy `data.*` field values for display without rewriting their persisted representation.
- Keep the existing column-clearing behavior when the user explicitly selects another Collection.

## How to Test

- `npm run run-cypress -- --spec tests/e2e/specs/CollectionRecordListColumns.spec.js --browser electron` — 6 passing.
- `npm run build-bundle` — passed.
- Manual validation on `processmaker.test` confirmed that the remapped Collection fixture and the zero-record Collection fixture preserve all six columns through Preview → Design, save, and reopen, while explicit Collection and Data Selection changes still clear the dependent configuration.

## Related Tickets & Packages

- https://processmaker.atlassian.net/browse/FOUR-32798
- https://processmaker.atlassian.net/browse/FOUR-32865
- Package: `@processmaker/screen-builder`

ci:deploy

### DIFF
--- a/src/components/inspector/collection-data-source.vue
+++ b/src/components/inspector/collection-data-source.vue
@@ -17,7 +17,7 @@
       <CollectionRecordsList
         v-model="collectionFields"
         :record-pmql="pmql"
-        @change="collectionChanged"
+        @collection-id-changed="collectionIdChanged"
       />
       <pmql-input
         v-model="pmql"
@@ -39,6 +39,7 @@
         v-model="dataSelectionOptions"
         :options="dataSelectionDisplayOptions"
         data-cy="inspector-collection-data-selection"
+        @change="dataSelectionChanged"
       />
       <small class="mt-3 form-text text-muted">
         {{ $t("The user can select specific data to be stored into a variable") }}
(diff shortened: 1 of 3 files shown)


nolanpro:
QA server K8S was successfully deployed https://ci-0f17f34122.engk8s.processmaker.net

eiresendez:
Hi @nolanpro I think that Screen Builder CI/CD appears unstable today: this PR failed across 3 attempts in unrelated Cypress specs, including identical blank-page failures in Computed Fields and Validation Rules, plus a separate File Download assertion failure. The PR-specific tests remain green, but these E2E failures caused coverage and scan to be skipped; the separate QA build, deployment, and PHPUnit workflow completed successfully.

CarliPinell:
The code looks good. No observations.

CarliPinell (approved):
No observations were found.

PR #1938, opened 2026-09-08
Bugfix/FOUR-32292-b: Validation of required fields does not work correctly in Record List controls with checkboxes

CarliPinell:
## Solution

Problems addressed

- Required Record Form fields were not validated correctly on parent Submit (values live in list rows, not at root).
- Add/Edit modal renderers polluted global valid/submitted state, so empty modal defaults could block parent Submit.
- Error message double-counted the same controls (vdata + schema / vocabularies → e.g. 4 fields → “8 validation errors”).
- PageNavigate on the Record Form page (e.g. “back to page 0”) pulled the main page into Record List row rules, creating ghost listName__* keys and inflating counts further (e.g. 10 vs 6).

Expected behavior after fix

- N required checkboxes (root + Record Form) → about N errors when invalid (not 2×N).
- Screens with PageNavigate on the Record Form page no longer inflate the count.
- Modal Add/Edit still validates locally; parent Submit validates list rows correctly.

## How to Test

Follow steps detailed in ticket

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-32292

## Code Review Checklist

ci:deploy

### DIFF
--- a/src/ValidationsFactory.js
+++ b/src/ValidationsFactory.js
@@ -137,7 +137,8 @@ class Validations {
    * Device visibility is decided once (screen-level). Conditional Visibility
    * Rules must NOT be applied here: they depend on live data and can change
    * per row / after the user toggles another control. Those expressions are
-   * evaluated inside each validator at submit/validate time instead.
+   * evaluated inside each validator at submit/validate time instead
+   * (field conditionalHide, parentVisibilityRule from containers / PageNavigate).
    */
   isVisible() {
     const visibleInDevice =
@@ -146,6 +147,21 @@ class Validations {
         : this.element.visibleInDevice;
     return !!visibleInDevice;
   }
+}
+
+/**
+ * AND-combine visibility expressions so nested parents and the current control
+ * both gate child validators at validate time.
+ */
+function combineVisibilityRules(...rules) {
+  const parts = rules.filter((rule) => typeof rule === 'string' && rule.trim());
+  if (parts.length === 0) {
+    return undefined;
+  }
+  if (parts.length === 1) {
+    return parts[0];
+  }
+  return parts.map((rule) => `(${rule})`).join(' and ');
 }
 
 /**
@@ -358,7 +374,19 @@ class PageNavigateValidations extends Validations {
     if (pagesValidated.length > 0 && !pagesValidated.includes(screenPageId)) {
       if (this.screen.config[screenNumber] && this.screen.config[screenNumber].items) {
         pagesValidated.push(screenPageId);
-        await ValidationsFactory(this.screen.config[this.element.config.eventData].items, { screen: this.screen, data: this.data }).addValidations(validations);
+        // Propagate the nav button's Visibility Rule (and any parent container
+        // rule) so destination required fields do not block submit while the
+        // page is unreachable. isVisible() no longer evaluates conditionalHide
+        // at build time, so this parentVisibilityRule is the runtime gate.
+        const parentVisibilityRule = combineVisibilityRules(
+          this.parentVisibilityRule,
+          this.element.config && this.element.config.conditionalHide
+        );
+        await ValidationsFactory(this.screen.config[this.element.config.eventData].items, {
+          screen: this.screen,
+          data: this.data,
+          parentVisibilityRule
+        }).addValidations(validations);
       }
     }
   }
--- a/src/ValidationsFactory.js
+++ b/src/ValidationsFactory.js
@@ -132,27 +132,19 @@ class Validations {
   }
 
   /**
-   * Check if element/container is visible.
+   * Check if element/container should contribute rules at build time.
+   *
+   * Device visibility is decided once (screen-level). Conditional Visibility
+   * Rules must NOT be applied here: they depend on live data and can change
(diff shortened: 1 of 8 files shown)


nolanpro:
QA server K8S was successfully deployed https://ci-016a15d52c.engk8s.processmaker.net

nolanpro:
QA server K8S was successfully deployed https://ci-016a15d52c.engk8s.processmaker.net

src/ValidationsFactory.js
```diff
@@ -235,6 +372,86 @@ class PageNavigateValidations extends Validations {
...
+      return;
+    }
+    const formPage = this.screen && this.screen.config && this.screen.config[formIndex];
+    if (!formPage || !Array.isArray(formPage.items)) {
+      return;
+    }
+
+    const rowRules = {};
+    const rows = get(this.data, listName);
+    const firstRow = (Array.isArray(rows) && rows.length > 0) ? rows[0] : {};
```
eiresendez: 🔴 
- [`src/ValidationsFactory.js:409`](https://github.com/ProcessMaker/screen-builder/blob/64d1bb0c202bf1f9350cdb57e6b5d995c0d24d03/src/ValidationsFactory.js#L409-L418) — Record Form validation rules depend on the first row’s visibility state.
  - The rule tree is constructed using `firstRow`. `FormElementValidations.isVisible()` can therefore omit a conditionally visible required field when it is hidden in the first row.
  - If that field is visible and invalid in a later row, no rule exists to validate it and the parent form can submit invalid data.
  - Build the validation rules independently of one representative row, while evaluating conditional visibility separately for every row during validation.

https://github.com/user-attachments/assets/3a074124-e28d-4232-aff2-5dc736fdd778

src/ValidationsFactory.js
```diff
@@ -235,6 +372,86 @@ class PageNavigateValidations extends Validations {
...
+      Object.keys(rules).forEach((ruleName) => {
+        const originalFn = rules[ruleName];
+        if (typeof originalFn !== 'function') {
+          return;
+        }
+        target[ruleName] = function(...props) {
+          const data = props[1];
+          const listRows = get(data, listName);
+          // No rows yet: treat as empty field values (required/accepted fail).
+          if (!Array.isArray(listRows) || listRows.length === 0) {
```
eiresendez, line 465: 🔴 The PR makes an empty Record List fail once for every required field in its Record Form. This behavior is explicitly covered by the new unit test, but it is outside the expected behavior of FOUR-32292.

@CarliPinell Could you confirm whether required fields inside the Record Form are also intended to require at least one row? Otherwise, this changes the previous behavior and may make existing screens with an optional empty Record List invalid.

eiresendez (changes requested):
Manual testing confirms that the reported scenario is fixed 👍  I'm just leaving a few observations:

1. Remove first-row-dependent conditional validation
2. Clarify empty-list semantics

src/ValidationsFactory.js
CarliPinell: For now the change is according QA requested. We assumed this is the expected behavior and product agrees.
But we can let it know to Product.

src/ValidationsFactory.js
CarliPinell: The observation was fixed.

nolanpro:
QA server K8S was successfully deployed https://ci-016a15d52c.engk8s.processmaker.net

src/ValidationsFactory.js
```diff
@@ -32,27 +132,19 @@ class Validations {
...
   /**
-   * Check if element/container is visible.
+   * Check if element/container should contribute rules at build time.
+   *
+   * Device visibility is decided once (screen-level). Conditional Visibility
+   * Rules must NOT be applied here: they depend on live data and can change
+   * per row / after the user toggles another control. Those expressions are
+   * evaluated inside each validator at submit/validate time instead.
    */
   isVisible() {
```
eiresendez, line 143: 🔴 `isVisible()` now ignores `conditionalHide` for every validation class, including `PageNavigateValidations`. A hidden navigation button will therefore still collect required rules from its destination page, but those validators do not inherit the button’s visibility condition. This can make an unreachable page block form submission. Please preserve conditional handling for PageNavigate, or propagate its visibility rule to the destination validators and add a regression test.

https://github.com/user-attachments/assets/840a98db-c148-4ed3-820a-24aab2e01196

eiresendez (changes requested):
hi @CarliPinell thanks for fixing the previous issue, the new commit introduced this new observation!

src/ValidationsFactory.js
CarliPinell: Thanks for the observation, the problem were solved. Please try again once the build environment is generated.

nolanpro:
QA server K8S was successfully deployed https://ci-016a15d52c.engk8s.processmaker.net

eiresendez (approved):
There are no remaining code-review blockers based on the diff and manual QA 👍👍

nolanpro:
QA server K8S was successfully deployed https://ci-016a15d52c.engk8s.processmaker.net

